PR #14, opened 2024-07-26
HCBS Header

rocio-desantiago:
### Description

- Header appears on Login
- Logout Button is functional.
- Updated icons for HCBS.

### Related ticket(s)

CMDCT-3755

---
### How to test

1. Login as a stateuser
2. Header should appear
Screenshot 2024-07-26 at 5 27 04 PM
3. Note: "Hello world" has been removed after some updates.
4. After merging main, expect to see Header and Footer smooshed "for now"
Screenshot 2024-07-30 at 3 37 15 PM

---
### Author checklist

---


rocio-desantiago:
`services/ui-src/src/components/menus/Menu.test.tsx` can be improved

services/ui-src/src/components/app/App.tsx
```diff
@@ -1,12 +1,15 @@
+import { useContext } from "react";
 import { Routes, Route } from "react-router-dom";
-import { LoginCognito, LoginIDM, PostLogoutRedirect } from "components";
-import { useStore } from "utils";
-import { makeMediaQueryClasses } from "utils/other/useBreakpoint";
-import { Container, Divider, Heading, Stack } from "@chakra-ui/react";
+// components
```
BearHanded, line 3: tiny nit: please remove these style comments in hcbs, we don't need to track the order here

ajaitasaini:
this would be a good time to add the Header test file in!

services/ui-src/src/components/layout/Header.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+      expect(screen.getByAltText("HCBS logo")).toBeVisible();
+    });
+
+    test("Help button is visible", () => {
+      expect(screen.getByAltText("Help")).toBeVisible();
+    });
+
+    test("Menu button is visible", () => {
+      expect(screen.getByAltText("Arrow down")).toBeVisible();
+    });
```
benmartin-coforma, line 33: I would probably merge these into one test; there isn't really a need to render the component 4 times to assert that it has 4 elements in it.
```suggestion
    test("Header renders correctly", () => {
      expect(screen.getByRole("navigation")).toBeVisible();
      expect(screen.getByAltText("HCBS logo")).toBeVisible();
      expect(screen.getByAltText("Help")).toBeVisible();
      expect(screen.getByAltText("Arrow down")).toBeVisible();
    });
```

services/ui-src/src/components/layout/Header.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+          text={"My Account"}
+          icon={"icon_arrrow_down.png"}
+          altText={"Arrow down"}
+        />
+      );
+
+      const menuButton = screen.getByRole("button", { name: /my account/i });
+      expect(menuButton).toBeInTheDocument();
+
+      fireEvent.click(menuButton);
```
benmartin-coforma, line 47: It's a good idea to use `userEvent` instead of `fireEvent` pretty much everywhere. `fireEvent` is a more low-level interface, whereas `userEvent` does a better job of simulating actual user flow.

It is in a separate package*, so if HCBS doesn't already have `@testing-library/user-event`, this would be a good time to install it.

*For now. The author (of both libraries) has said he intends to eventually include `userEvent` in the main package. For this and more, see his blog: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event

```suggestion
      userEvent.click(menuButton);
```

benmartin-coforma: This test is also missing any kind of assertion about what happens when the menu button is clicked. Will that cause a navigation? We can detect those in unit tests. Will it cause another element to become visible? We can detect that too.

services/ui-src/src/components/layout/Header.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+          text={"Log Out"}
+          icon={"icon_arrow_right_square.png"}
+          altText={"Logout"}
+        />
+      );
+
+      const logoutButton = screen.getByRole("img", { name: /Logout/i });
+      expect(logoutButton).toBeInTheDocument();
+
+      fireEvent.click(logoutButton);
```
benmartin-coforma, line 62: Same as above - let's use `userEvent`, and check to see that something has happened after the button is clicked.

services/ui-src/src/components/menus/Menu.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+const menuComponent = (
+  <RouterWrappedComponent>
+    <Menu handleLogout={() => {}} />
+  </RouterWrappedComponent>
+);
+
+describe("<Menu />", () => {
+  test("Menu button is visible", () => {
+    render(menuComponent);
+    expect(screen.getByTestId("header-menu-dropdown-button")).toBeVisible();
```
benmartin-coforma, line 15: I _think_ you'll be able to find this element by `aria-label` instead of `data-testid`. Test ID should generally be a last resort for locating elements in tests, because we're trying to emulate how users interact with the page. They can see (or hear) text and labels, but they'll never check the DOM for test IDs.

services/ui-src/src/components/menus/Menu.tsx
```diff
@@ -0,0 +1,107 @@
...
+    <MenuRoot offset={[8, 20]}>
+      <Box role="group">
+        <MenuButton
+          as={Button}
+          rightIcon={
+            <Image src={chevronDownIcon} alt="Arrow down" sx={sx.menuIcon} />
+          }
+          sx={sx.menuButton}
+          aria-label="my account"
+          data-testid="header-menu-dropdown-button"
```
benmartin-coforma, line 31: _If_ we can get the tests going without querying for Test IDs, we should strip them out of the markup as well.

```suggestion
```

benmartin-coforma (approved):
Looks good to me! I have some nitpicky comments below, but I'd also be fine merging this as-is.

services/ui-src/src/components/menus/Menu.test.tsx
benmartin-coforma: I do recommend reading through the "Which Query Should I Use" page of the Testing Library docs: https://testing-library.com/docs/queries/about#priority

rocio-desantiago: Oooh good tidbit. I've noticed we frequently use `getbyTestId` throughout MFP. Now's a good time to address this across our initial tests while HCBS is still fresh :D

services/ui-src/src/components/layout/Header.test.tsx
rocio-desantiago: Thank you for that! While researching how to make test more effective, I ran into the `fireEvent` but `userEvent` is definitely the better option here.

services/ui-src/src/components/layout/Header.test.tsx
rocio-desantiago: Good idea. There may be other places to refine the tests like this too.

services/ui-src/src/components/menus/Menu.tsx
rocio-desantiago: We do use `data-testid` a lot. I wonder why that became a common pattern.

services/ui-src/src/components/layout/Header.test.tsx
rocio-desantiago: Oooh just realized that's the same thing that 
`expect(logoutButton).toBeInTheDocument();`
is doing, per the suggestion of using 
`expect(screen.getByText("Log out")).toBeInTheDocument();`

Leaving as-is.

BearHanded (approved):

ajaitasaini (approved):

PR #32, opened 2024-09-11
[HCBS] Create Submission Dashboard

ajaitasaini:
### Related ticket(s)

CMDCT-3948

---
### How to test

make sure you can see this page 
image

### Important updates

---
### Author checklist

---


services/ui-src/src/components/accordions/InstructionsAccordion.tsx
```diff
@@ -0,0 +1,75 @@
+// components
```
BearHanded, line 1: remove import comments

BearHanded (approved):

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
benmartin-coforma, line 1: Comment: This file contains a fair bit that we can simplify for HCBS. For example, the `getStatus()` function is likely to change, and the table content filtering seems MFP-specific. That said, I'm fine merging it in as-is for now; we can simplify later when we have a more specific idea of how this will work.

services/ui-src/src/components/accordions/InstructionsAccordion.tsx
```diff
@@ -0,0 +1,72 @@
...
+        )
+      )}
+    </UnorderedList>
+  );
+};
+
+interface Props {
+  verbiage: AnyObject;
+  [key: string]: any;
+}
```
benmartin-coforma, line 36: Thought: I'd love to see something more specific here.
* We know that the verbiage can only have 4 properties, 2 of which are required (buttonLabel, intro, list, text). So we could list those here to get better type safety than `AnyObject`.
* We know that there are a finite number of properties that are acceptable to pass through to Chakra's `Accordion`. We could restrict our props to only those by having `Props extends AccordionProps { ... }`
   * With AccordionProps being imported from Chakra
   * And with the `[key: string]: any` removed from our interface
   * We probably don't want the `allowToggle`, `allowMultiple`, or `sx` props overridden, so we could exclude those using `Omit` ( https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys )
* We know that, in all our other apps, the only prop we ever override on Chakra's Accordion is `defaultIndex={0}` - and we always need a comment explaining why. So perhaps we could skip the passthrough `...props` altogether, while removing the need for those explanations, by introducing a new named prop: `isOpenByDefault?: boolean`

Buuut this super-loose `Props` interface isn't actively harming us, either, so there's no need to address it as a part of this PR.

services/ui-src/src/components/accordions/InstructionsAccordion.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <AccordionItem label={buttonLabel} sx={sx.item}>
+        <Box sx={sx.textBox}>{parseCustomHtml(intro)}</Box>
+        {list && parseList(list)}
+        {text && <Box sx={sx.textBox}>{parseCustomHtml(text)}</Box>}
+      </AccordionItem>
+    </Accordion>
+  );
+};
+
+export const parseList = (list: any) => {
```
benmartin-coforma, line 19: Thought: the strictest way to type this parameter (instead of `any`) would be with a recursive type, like this:
```ts
type NestedStringArray = (string | NestedStringArray[])[];
export const parseList = (list: NestedStringArray) => { ... }
```
And in fact, I doubt we would ever (or _should_ ever) have a triply-nested unordered list inside of an accordion, so we could probably get away with `(string | string[])[]`.

But, once again the `any` isn't actively hurting anybody, so there's no rush to change it in the context of this PR.

benmartin-coforma (changes requested):
I'm requesting changes for `rendering.ts` - the rest of my comments are more idle musing.

services/ui-src/src/components/pages/Dashboard/DashboardPage.test.tsx
```diff
@@ -0,0 +1,62 @@
+import { render, screen } from "@testing-library/react";
+// components
```
rocio-desantiago, line 2: we can remove import comments here

rocio-desantiago (changes requested):
lingering import comments to be removed. Do we need to keep new lines in the import section?

benmartin-coforma (approved):

rocio-desantiago (approved):

PR #33, opened 2024-09-18
[Spike] Interactive QM Report

BearHanded:
### Description

The goal of this PR is to lay the foundation for continued iteration. It aggregates efforts made by multiple developers and it attempts to accomplish the following:
- Setup a report template structure
- Add database service, add qm table
- Potential report template structure
- Create and get a report on interaction
- Navigate the above structure
- Fields can be interacted with
- Adds the storage pattern of db interaction

It falls short as follows:
- Only the Textbox element is writing to react hook form and store. <-
- Writing to react hook store is going into an answers array, which gets slapped onto the report object. Original intention was to have the item saved as an answer in the object itself, right alongside the question text.
  - If this is impossible, it might make sense to just wrap the measure sections back to unduplicated templates
- It does not post the object to be saved.
- Test Coverage

### Related ticket(s)

CMDCT-3761

---
### How to test

- Run the app || https://d2a3efhesaor7a.cloudfront.net/
- Log in as state user
- Navigate to the QM report
- The page that gives you options for report setup will actually create an object on submit
- The report view actually pulls that created item back from the api
- The page can be navigated with sidebar
- The required and stratified views filter in measures.
- The first edit button pops up a modal, that doesn't do much at the moment
- The second links you to a measure page
- You can return from the measure page

### Important updates

Update env file - new db table var.

---
### Author checklist

---


ailZhou (approved):
The code looks great! Feels like a real structure is in place compare to the ideas we had floating around.

There are some UI pieces i would like cleaned up (i.e. sidebar, matching components to the mockup) but i can do those.

angelaco11 (approved):
nice work!!!

PR #36, opened 2024-10-03
Text area field

ajaitasaini:
### Description

This PR adds the TextArea component to HCBS

### Related ticket(s)

CMDCT-3999

---
### How to test

Deploy link: https://d3axvrw92liyjc.cloudfront.net/
Navigate to the general information page, confirm that the first two text areas look stylized, confirm that you can enter information into the text boxes (the text areas do not need to save any information you're inputting, that will follow in a later ticket)

### Important updates

---
### Author checklist

---


services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  // get form context and register field
+  const form = useFormContext();
+  const inputName = "answers.1.2"; // TODO: id based
+
+  useEffect(() => {
+    form.register(inputName);
+  }, []);
+
+  const onChangeHandler = async (
```
BearHanded, line 22: Does this work with onBlur only or does it need onChange? My concern is frequency of updates being shuffled up to api calls

ajaitasaini: my understanding was that both are eventually needed (if they click next without blurring from the field)

benmartin-coforma: I think clicking Next _will_ blur from the field, and `onChange` fires every keystroke.

If we want field validation on keystroke, then we need an onChange handler. If we want to limit API calls, we should only send data back to the API onBlur.

In our other apps, onBlur is when we would call autosave code. If HCBS automatically percolates store updates to the API, and we're sending data to the store on every keystroke, maybe we could do some kind of debounce?

Alternatively, perhaps we could divorce store updates from field validation somehow? I'm not sure how that would work, though...

benmartin-coforma (dismissed):
I think this is a good starting point, and we should get it merged in. If we decide to make changes to the field's interaction with the store, we can do that in future PRs.

benmartin-coforma (approved):

BearHanded (approved):

PR #40, opened 2024-10-15
CMDCT-3958: Swap to OAC on UI Bucket

BearHanded:
### Description

SecurityHub is mad because we don't use the 2022 new and shiny Origin Access Control rules for our cloudfront distro. Instead we are using Origin Access Identity, which is a slightly more complex pattern to do the same thing.

I am pretty sure the old CMS AWS accounts have this rule disabled, and HCBS is flagged because it is under the new account setup. Regardless, easy enough to clean up and remove a slightly more complicated config in favor of one that just sets up a single rule between two resources.

This just takes the bucket we host our UI in, and says it can only be accessed via the specific cloudfront distribution. No public access by navigating directly to the bucket.

### Related ticket(s)

CMDCT-3958

---
### How to test

https://du8ioa447dep0.cloudfront.net/
Deployed link still works, you wouldn't be able to access the bucket if things went really wrong.
~~I will monitor securityhub for the next scan for this branch~~
Confirmed this resolves the finding.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui/serverless.yml
+++ b/services/ui/serverless.yml
@@ -143,6 +143,11 @@ resources:
           ServerSideEncryptionConfiguration:
             - ServerSideEncryptionByDefault:
                 SSEAlgorithm: AES256
+        PublicAccessBlockConfiguration:
+          BlockPublicAcls: true
+          BlockPublicPolicy: true
+          IgnorePublicAcls: true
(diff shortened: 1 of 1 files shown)


ailZhou (approved):
ran deploy with no issues

PR #41, opened 2024-10-21
[Cmdct-3998] Save and Load Form

BearHanded:
### Description

Introduce a light version of autosave and populate the modified form when revisiting the page.

- Merges the answers from the form back into the report, then submits.
- Only triggers that update onBlur, the typing in each field is a separate event that still updates the field.
- Each input object is taking its "formkey" (its path in the json from the page level), and applying "answer" itself. More complex inputs or input groups can follow the same pattern to declare their own field in the json, or structure.
- The report is loaded into react hook form to provide defaults. By virtue of these paths matching, it pre-populates without the hydration logic that is baked into MFP & MCR. 
- Moves PageMap to a dictionary of <Key, Number> where it just keeps track of the index. We can avoid tracking two different copies of the pages and just have a quick lookup. We can further bake the lookups into a util file if they keep recurring.

Open Questions:
- Triggering the post from the action felt like the quickest path to success, but seems like it is a coin flip whether folks say it is good practice or not. We can handle writing to an error state from there if something goes wrong.
- Validation. The lifecycles of onBlur and onChange seem separate, but we'll need to find how we plug it into each field still. Each element component should have the info it needs to get it if we have rules or a rules lookup declared at its level, but more complex validation might be a curveball.

### Related ticket(s)

CMDCT-3998

---
### How to test

https://dr2j580shd4xj.cloudfront.net/
Login as a state user.
- When you edit a form, and blur, the network tab shows submission of the data
- When you change pages, that data persists
- When you refresh the page, that data persists

Other things touched:
- PageMap lookups. Sidebar and Previous/Continue buttons should continue to work as expected.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/report/StatusTable.test.tsx
+++ b/services/ui-src/src/components/report/StatusTable.test.tsx
@@ -19,18 +19,27 @@ describe("StatusTableElement", () => {
   beforeEach(() => {
     (useNavigate as jest.Mock).mockReturnValue(mockNavigate);
     jest.clearAllMocks();
+
+    const mockPageMap = new Map();
+    mockPageMap.set("root", 0);
+    mockPageMap.set("1", 1);
+    mockPageMap.set("2", 2);
+
+    const report = {
+      pages: [
+        { childPageIds: ["1", "2"] },
+        { title: "Section 1", id: "id-1" },
+        { title: "Section 2", id: "id-2" },
+      ],
+    };
+    (useStore as unknown as jest.Mock).mockReturnValue({
+      pageMap: mockPageMap,
+      report: report,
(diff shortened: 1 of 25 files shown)


ailZhou (approved):
lgtm! tested in deploy, data is saving/loading when i change pages

services/ui-src/src/components/report/StatusTable.test.tsx
rocio-desantiago, line 1: Nice clean up!

rocio-desantiago (approved):
Awesome refactoring and awesome updates to the tests! 
Tested in deploy link and working as expected ✅

PR #48, opened 2024-10-28
Add authorization to the API

benmartin-coforma:
### Description
This PR adds permission checks to the API, but it also restructures it.

Currently in main, the API flow looks like this:
handler_before
1. handler-lib verifies the signature on the JWT, confirming that this request is from a valid user
2. handler-lib calls the appropriate lambda, passing in the entire Event
3. The lambda reads the path parameters, such as state and reportType, from the Event
4. The lambda reads the user data from the JWT
   * This includes a one-size-fits-all permission check; that either the user is an admin or they are from the correct state
5. The lambda does whatever actual operation it's supposed to

This PR changes the flow, to this:
handler_after
1. handler-lib verifies the signature on the JWT _while also_ reading the user data
2. handler-lib reads the path parameters from the Event
   * Making use of whichever Parameter Parser it receives from the endpoint
3. handler-lib calls the appropriate lambda, passing in the refined Request
   * (I apologize if the layout of the diagram makes the sequence unclear)
5. The lambda performs whichever permission checks are appropriate
6. The lambda does whatever actual operation it's supposed to

I will admit this feels a little awkward, with the parser being passed up from the endpoint to handler-lib. I'll also note that this is among the first uses of a generic type parameter in our codebase: The return type of the parser (such as `parseReportParameters`) flows through to determine the type details of the Request (such as the presence/absence of `request.parameters.id`).

***Question for Reviewers:*** Is this tradeoff worth it? What we gain is the removal of a couple lines of parameter-parsing boilerplate, at the top of every endpoint lambda. What we pay is this increase in complexity. Is the API flow still easy enough to follow? Will the generic type parameters be a hurdle for future maintainers?

### Related ticket(s)
CMDCT-4005

---
### How to test
1. Log in with `stateuser@test.com` at https://d1x4yjcaf95sje.cloudfront.net/
2. Create a report - note that it succeeds
3. Make some changes to the report data - note that it saves successfully
4. Go back to the QM dashboard, and open the browser console
5. Change the state in the URL, and navigate there
7. Note the error in the console. This came from our new authorization check.

### Important updates
This PR moves parameter parsing to handler-lib! This is a major pattern change, for the whole API.

---
### Author checklist


services/app-api/libs/handler-lib.ts
benmartin-coforma, line 1: A few extra little changes to note in here:
1. The try/catch now encompasses the entire body of `handler()`
   * This ensures that all but the most catastrophic errors will be logged in our logger
   * Which includes any errors that occur during user authentication, parameter parsing, and so on
2. The function is de-nested overall by moving `return unauthenticated()` from an `else` to an early return.
3. The second argument of `handler()` is gone. We never used it.
   * I believe AWS will continue to pass it, and we will still ignore it without issue.
   * JS cares very little for function [arity](https://en.wikipedia.org/wiki/Arity), as this shows:
   ```js
   const add1 = x => x + 1;
   console.log(add1(5)) //=> 6
   console.log(add1(5, "an extra arg")) //=> still just 6, with no problems
   ```

services/app-api/libs/param-lib.ts
benmartin-coforma, line 1: I had to make a choice here, on how to represent parse failures. Options include:
* Throw an Error, to be caught by the handler
   + Pro: Any caller is forced (by the runtime) to deal with failures
   + Pro: We can deliver a more specific message to the user
      - Example parser code: `throw new Error("Must supply a valid report type");`
      - Example handler code: `return badRequest(err.message);`
   + Con: All calls must be wrapped in try/catch
   + Con: Typescript will not remind you if you forget the try/catch
* Return undefined (which is where I landed)
   + Pro: Any caller is forced (by Typescript) to deal with failures
   + Con: A somewhat arbitrary magic value
* Return an object containing a success flag (which is what we're changing from)
   + Example call: `const { success, ...params } = parseReportParameters()`
   + Pro: Typescript support
   + Con: All of these `as const` statements are noisy
   + Con: The type definition of AuthenticatedRequest becomes more complicated
      - (or alternatively, `AuthenticatedRequest.parameters` always includes this flag)
* Return an object containing a (possibly-empty) list of issues, and a sub-object of the parameters
   + Example call:
   ```ts
   const { issues, params } = parseReportParameters()
   if (issues.length > 0) {
     return badRequest(`Invalid request parameter(s): ${issues.join("; ")}`);
   }
   ```
   + Pro: The parameter parser no longer has a need to interact directly with the logger
   + Pro: More detailed messages to the users
   + Pro: Can report on multiple invalid parameters at once
   + Con: Typescript support is lost
      - (or alternatively, some _real_ type gymnastics to distinguish empty arrays from nonempty ones)

In the end, I ruled out Errors because Typescript support is too valuable. I was tempted by the issue list because of the extra detail we could potentially get on the UI side, but decided that wasn't worth it either. The most likely time for an error like "invalid report type" is during development, when an engineer is adding a new report type, and they will have easy access to the app logs for troubleshooting.

Between `undefined` and some sort of success flag, I decided that the simplicity of `undefined` wins out. It is somewhat arbitrary &mdash; for example, the return type could be `User | false` instead &mdash; but it shouldn't ever be _too_ surprising. I would bet it's the [Schelling Point](https://en.wikipedia.org/wiki/Focal_point_(game_theory)) of functions like these.

services/app-api/utils/authentication.ts
```diff
@@ -0,0 +1,100 @@
...
+  }
+  return state;
+};
+
+const parseEmailFromToken = (token: CognitoIdTokenPayload) => {
+  if (!token.email_verified) {
+    throw new Error(`User's email is not verified`);
+  }
+  return token.email as string;
+};
```
benmartin-coforma, line 94: We should probably go through our various Cognito pools and confirm that all users have verified emails. If they do not, this check should be removed.

In fact, given the risk of user confusion, perhaps we should remove this check regardless.

benmartin-coforma:
The change from APIGatewayProxyEvent to AuthenticatedRequest was inspired by this blog post: [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)

services/app-api/utils/authentication.ts
BearHanded: this is my only open question before approval, the rest looks like good changes

rocio-desantiago: Do we want to go ahead and remove `parseEmailFromToken`? since Brendan confirmed emails are not verified?

rocio-desantiago (commented):
question about email verification

rocio-desantiago:
Oops there's a merge conflict with `services/app-api/handlers/reports/update.ts`

services/app-api/utils/authentication.ts
benmartin-coforma: Done.

rocio-desantiago (approved):
👍

ajaitasaini (changes requested):
when changing the state in the URL from PR to NJ, there was indeed a console error but it technically did navigate me to a New Jersey reports table UI. is that expected for now?

benmartin-coforma:
> when changing the state in the URL from PR to NJ, there was indeed a console error but it technically did navigate me to a New Jersey Quality Measures Report table UI. is that expected for now, or should I be seeing an error page as well?

This is expected for now, and the foreseeable future. Users manually navigating to other state's dashboards is a scenario we need to _securely_ handle, but not one we need to _gracefully_ handle. The page will show the state's name (because that's a UI-side lookup in a cosmetic abbr->name map), but it will not show any reports or report data for that state.

BearHanded (approved):

PR #51, opened 2024-10-29
[HCBS] PRA Disclosure

rocio-desantiago:
### Description

**Other ideas:**
- Initially, I would have loved to add the PRA disclosure statement as an `ElementType.Paragraph` to our report template file: `qm.ts` but the logic to show that after the button was getting a little quirky. 
- Another great idea would to keep it in its own component, and have it display if the report page has some flag like `showPraDisclosure`
- There's also the question of whether that supports the idea of creating a `ReportPageFooter` component yet.

**Present Solution**
- Considering that this is the only content for under the buttons as of now, the most direct way was adding it to the bottom of `ReportPageWrapper`. 
- After some feedback, TODO for Divider when Tables appear on page since the lines look like they're doubled due to the bottom of the page, `ParagraphTempate` moved into `constants` file instead of `qm.ts` file.

**Before feedback:**
> - I also moved the divider because on the following pages, the divider was looking to appear twice due to the existing divider from the tables on those pages.
> - Added a separate `ParagraphTemplate` in the `qm.ts` file but this feels hacky. Open to feedback.

### Related ticket(s)

CMDCT-4001

---
### How to test

Deploy Link: https://d2dkwx0584k7ea.cloudfront.net/
1. Login as a state user to view a Quality Measures Report.
2. You should see the PRA disclosure content under **General Information** tab.
3. Tab through the pages and confirm it only shows up under General Information.
  - Privacy Rights Act Disclosure

---
### Author checklist

---

### DIFF
--- a/services/app-api/forms/qm.ts
+++ b/services/app-api/forms/qm.ts
@@ -4,6 +4,7 @@ import {
   ElementType,
   ReportType,
   MeasureTemplateName,
+  ParagraphTemplate,
 } from "../types/reports";
 
 export const qmReportTemplate: ReportTemplate = {
@@ -230,4 +231,10 @@ export const qmReportTemplate: ReportTemplate = {
       ],
     },
   },
+};
+
+export const praStatement: ParagraphTemplate = {
+  type: ElementType.Paragraph,
+  title: "PRA Disclosure Statement",
+  text: "Under the Privacy Act of 1974 any personally identifying information obtained will be kept private to the extent of the law. According to the Paperwork Reduction Act of 1995, no persons are required to respond to a collection of information unless it displays a valid OMB control number. The valid OMB control number for this information collection is 0938-1053. The time required to complete this information collection is estimated to average 2.5 hours per response, including the time to review instructions, search existing data resources, gather the data needed, and complete and review the information collection. If you have comments concerning the accuracy of the time estimate(s) or suggestions for improving this form, please write to: CMS, 7500 Security Boulevard, Attn: PRA Reports Clearance Officer, Mail Stop C4-26-05, Baltimore, Maryland 21244-1850",
 };
--- a/services/app-api/forms/qm.ts
+++ b/services/app-api/forms/qm.ts
@@ -4,7 +4,6 @@ import {
   ElementType,
   ReportType,
   MeasureTemplateName,
-  ParagraphTemplate,
 } from "../types/reports";
 
 export const qmReportTemplate: ReportTemplate = {
@@ -231,10 +230,4 @@ export const qmReportTemplate: ReportTemplate = {
       ],
     },
   },
-};
-
-export const praStatement: ParagraphTemplate = {
-  type: ElementType.Paragraph,
-  title: "PRA Disclosure Statement",
-  text: "Under the Privacy Act of 1974 any personally identifying information obtained will be kept private to the extent of the law. According to the Paperwork Reduction Act of 1995, no persons are required to respond to a collection of information unless it displays a valid OMB control number. The valid OMB control number for this information collection is 0938-1053. The time required to complete this information collection is estimated to average 2.5 hours per response, including the time to review instructions, search existing data resources, gather the data needed, and complete and review the information collection. If you have comments concerning the accuracy of the time estimate(s) or suggestions for improving this form, please write to: CMS, 7500 Security Boulevard, Attn: PRA Reports Clearance Officer, Mail Stop C4-26-05, Baltimore, Maryland 21244-1850",
 };
--- a/services/ui-src/src/components/report/ReportPageWrapper.tsx
+++ b/services/ui-src/src/components/report/ReportPageWrapper.tsx
@@ -8,10 +8,9 @@ import { useParams } from "react-router-dom";
 import { useStore } from "utils";
 import { FormProvider, useForm } from "react-hook-form";
 import { FormPageTemplate } from "types/report";
+import { praStatement } from "../../../../app-api/forms/qm";
 
 export const ReportPageWrapper = () => {
-  const praStatement =
-    "Under the Privacy Act of 1974 any personally identifying information obtained will be kept private to the extent of the law. According to the Paperwork Reduction Act of 1995, no persons are required to respond to a collection of information unless it displays a valid OMB control number. The valid OMB control number for this information collection is 0938-1053. The time required to complete this information collection is estimated to average 2.5 hours per response, including the time to review instructions, search existing data resources, gather the data needed, and complete and review the information collection. If you have comments concerning the accuracy of the time estimate(s) or suggestions for improving this form, please write to: CMS, 7500 Security Boulevard, Attn: PRA Reports Clearance Officer, Mail Stop C4-26-05, Baltimore, Maryland 21244-1850";
   const {
     report,
     pageMap,
@@ -126,12 +125,7 @@ export const ReportPageWrapper = () => {
               </Flex>
               <Box flex="auto" width="97.5%">
                 {parentPage.index == 0 && (
-                  <>
-                    <h4>
-                      <b>PRA Disclosure Statement</b>
-                    </h4>
-                    <p>{praStatement}</p>
-                  </>
+                  <Page elements={[praStatement] ?? []}></Page>
                 )}
               </Box>
             </>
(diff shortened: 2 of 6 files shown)


services/ui-src/src/components/report/ReportPageWrapper.tsx
```diff
@@ -103,15 +105,35 @@ export const ReportPageWrapper = () => {
...
                   <Button
                     onClick={() => SetPageIndex(parentPage.index + 1)}
                     marginLeft="auto"
+                    justifyContent={"flex-end"}
                   >
                     Continue
                   </Button>
                 )}
               </Flex>
+              <Box flex="auto" width="97.5%">
```
rocio-desantiago, line 127: What's the consensus here on half percentages? This was the perfect number to get the text to match and the words to align as shown in [Figma](https://www.figma.com/design/B8IEoYKQ7DuQyx5nStls8F/HCBS?node-id=1-2&node-type=canvas&t=ikaqfqzeKN7Uki1Y-0).

services/ui-src/src/components/report/ReportPageWrapper.tsx
```diff
@@ -10,6 +10,8 @@ import { FormProvider, useForm } from "react-hook-form";
 import { FormPageTemplate } from "types/report";
 
 export const ReportPageWrapper = () => {
+  const praStatement =
+    "Under the Privacy Act of 1974 any personally identifying information obtained will be kept private to the extent of the law. According to the Paperwork Reduction Act of 1995, no persons are required to respond to a collection of information unless it displays a valid OMB control number. The valid OMB control number for this information collection is 0938-1053. The time required to complete this information collection is estimated to average 2.5 hours per response, including the time to review instructions, search existing data resources, gather the data needed, and complete and review the information collection. If you have comments concerning the accuracy of the time estimate(s) or suggestions for improving this form, please write to: CMS, 7500 Security Boulevard, Attn: PRA Reports Clearance Officer, Mail Stop C4-26-05, Baltimore, Maryland 21244-1850";
```
ajaitasaini, line 14: is there any way we could move this to the general information content and access it from there? I think the PRA statement only exists for the General Information Page anyway, while the ReportPageWrapper is used generally for all report pages

rocio-desantiago: Oh I was overthinking of how to make that work but yes I can move this!

services/ui-src/src/components/report/ReportPageWrapper.tsx
```diff
@@ -103,15 +104,30 @@ export const ReportPageWrapper = () => {
...
                     marginLeft="auto"
+                    justifyContent={"flex-end"}
                   >
                     Continue
                   </Button>
                 )}
               </Flex>
+              <Box flex="auto" width="97.5%">
+                {parentPage.index == 0 && (
+                  <Page elements={[praStatement] ?? []}></Page>
```
rocio-desantiago, line 128: This feels hacky and I'm open to feedback if there's a better way to pull the pra disclosure statment?

services/ui-src/src/components/report/ReportPageWrapper.tsx
rocio-desantiago: @ajaitasaini Moved! ✅

services/ui-src/src/components/report/ReportPageWrapper.tsx
BearHanded: Should this be a set margin instead? Might be worth checking

services/ui-src/src/components/report/ReportPageWrapper.tsx
```diff
@@ -103,15 +104,30 @@ export const ReportPageWrapper = () => {
                     Previous
                   </Button>
                 )}
-                {parentPage.index < parentPage.childPageIds.length - 1 && (
+                {parentPage.index < parentPage.childPageIds.length - 1 &&
```
BearHanded, line 107: Why are there 2 different ways of rendering the continue button?  Is it just to change the justifyContent? If so that can be handled with logic rather than duplicating the component

services/ui-src/src/components/report/ReportPageWrapper.tsx
BearHanded: It shouldn't need to be its own page, just an element right? Unless we're doing something where we support many elements, but it looks like it is just the PRA element.

services/app-api/forms/qm.ts
```diff
@@ -231,3 +232,9 @@ export const qmReportTemplate: ReportTemplate = {
     },
   },
 };
+
+export const praStatement: ParagraphTemplate = {
```
BearHanded, line 236: There might be a few assumptions being made here. By saving it alongside our qm template we are implying it needs to be maintained at the same frequency, and uniquely for a report.

Rather my guess is this disclosure might be safe to maintain in a single place for the entire application, and may update for every report that exists, rather than a single report. Can you clarify this with Cam and Nat?

services/ui-src/src/components/report/ReportPageWrapper.tsx
BearHanded: This also may render the pra statement on nested sets of pages, I believe this is the index within the parent's children

services/ui-src/src/components/report/ReportPageWrapper.tsx
rocio-desantiago: Removed. Per convo with Aileen

services/ui-src/src/components/report/ReportPageWrapper.tsx
rocio-desantiago: Good catch! There's no need for the extra logic. It's handled with the first one now.

services/ui-src/src/components/report/ReportPageWrapper.tsx
```diff
@@ -93,7 +94,7 @@ export const ReportPageWrapper = () => {
           </Box>
           {!currentPage.hideNavButtons && parentPage && (
             <>
-              <Divider></Divider>
+              {parentPage.index == 0 && <Divider></Divider>}
```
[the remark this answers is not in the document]
rocio-desantiago: Visually, only the first page needs a `<Divider>` because the rest of the pages have what looks like a Divider already because they're tables. So leaving the `<Divider>` in as it was before makes it appear as double lines - not matching the Figma design. 

I'm unclear on how sustainable this approach is considering the nested pages may not always be tables if we add more pages to other reports in the future.

| Required Measure Results | Stratified Measure Results |
| ------- | ------ |
| table 1 | table 1 |

services/app-api/forms/qm.ts
rocio-desantiago: Confirmed with Cam that for now we can assume that we might re-use it for other reports. I moved the content to the existing `constants.ts` file in `app-api/utils`

services/ui-src/src/components/report/ReportPageWrapper.tsx
rocio-desantiago: Correct. No need to pull it in via the `Page`. Pulling it in with the `paragraphElement` function but not sure about `index` and `formKey` props. Since design confirmed that we can assume other reports will also have this paragraph, are we good to keep it here?

BearHanded: `formKey` should is used for filling out answers, it might be fine to make an optional prop or exclude from a paragraph, or we make some other element that renders a block of text to the screen. Why not just make a PRADisclosure element and drop that here instead?

services/ui-src/src/components/report/ReportPageWrapper.tsx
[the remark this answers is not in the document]
BearHanded: I'm guessing the Divider here should be mandatory and elments like tables should skip it if they're last. Regardless, can you add a TO DO comment here, and a ticket to solidify the Divider behavior for our form controls vs elements in a page? It feels like our discussion here will scope creep otherwise.

services/ui-src/src/components/report/ReportPageWrapper.tsx
rocio-desantiago: ** PRADisclosure** element felt the most straightforward for now 😺

services/ui-src/src/components/report/ReportPageWrapper.tsx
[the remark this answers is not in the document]
rocio-desantiago: Done ✅

angelaco11 (dismissed):
tested and everything looks good

BearHanded (approved):

angelaco11 (approved):

PR #58, opened 2024-11-13
Admin Banner

ailZhou:
### Description

Behold, we have admin banner page & form!

Screenshot 2024-11-18 at 3 39 53 PM

### Related ticket(s)

CMDCT-4089
CMDCT-4090

---
### How to test

1) Sign into HCBS, [adminuser@test.com](mailto:adminuser@test.com)
2) In the top right corner, click [My Account] -> [Manage Account]
Screenshot 2024-11-18 at 3 33 32 PM
3) On the **My Account** page click [Banner Editor]
Screenshot 2024-11-18 at 3 34 27 PM
4)  This will take you to the **Banner Admin** page. Fill out the form completely and click [Replace Current Banner]
    - **Note**: We do not have validation yet so there will still be some bugginess until that is added
Screenshot 2024-11-18 at 3 36 49 PM
5) You should see the Banner filled in **Current Banner** section
Screenshot 2024-11-18 at 3 38 54 PM
6) Click [Delete Current Banner] and that should remove the banner. That should complete all the test for the banner.

### Important updates

---
### Author checklist

---


services/ui-src/src/components/banners/PreviewBanner.tsx
```diff
@@ -0,0 +1,23 @@
+import { useFormContext } from "react-hook-form";
+// components
```
benmartin-coforma, line 2:
```suggestion
```

services/ui-src/src/components/banners/PreviewBanner.tsx
```diff
@@ -0,0 +1,23 @@
...
+    description:
+      formData?.["bannerDescription"]?.["answer"] || "New banner description",
+    link: formData?.["bannerLink"]?.["answer"] || "",
+  };
+
+  return <Banner bannerData={bannerData} {...props} />;
+};
+
+interface Props {
+  [key: string]: any;
```
benmartin-coforma, line 22: If there are no properties in the props, would we just omit the argument on line 5 entirely? I think that works, right? 🤔 

```suggestion
```

services/ui-src/src/components/forms/AdminBannerForm.test.tsx
```diff
@@ -0,0 +1,94 @@
+import { render, screen } from "@testing-library/react";
+// components
```
benmartin-coforma, line 2:
```suggestion
```

services/ui-src/src/components/forms/AdminBannerForm.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { AdminBannerForm } from "components";
+// utils
```
benmartin-coforma, line 4:
```suggestion
```

services/ui-src/src/components/forms/AdminBannerForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+              element={{
+                type: ElementType.Textbox,
+                label: "Title text",
+              }}
+              formkey={"bannerTitle"}
+            ></TextField>
+            <TextField
+              element={{
+                type: ElementType.Textbox,
+                label: "Description tex",
```
benmartin-coforma:
```suggestion
                label: "Description text",
```

services/ui-src/src/components/forms/AdminBannerForm.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    />
+  </RouterWrappedComponent>
+);
+
+describe("<AdminBannerForm />", () => {
+  test("AdminBannerForm is visible", () => {
+    render(adminBannerFormComponent(mockWriteAdminBanner));
+    const form = screen.getByTestId("test-form");
+    expect(form).toBeVisible();
+  });
```
benmartin-coforma, line 26: I prefer to avoid `getByTestId` whenever possible, usually in favor of `getByRole`. In this case, there is no ARIA "form" role, so an equivalent test would have to check the various elements within the form, to ensure they are all visible. Something like
```ts
    expect(screen.getByRole("textbox", { name: "Title text"} })).toBeVisible();
    expect(screen.getByRole("textbox", { name: "Description text"} })).toBeVisible();
    expect(screen.getByRole("textbox", { name: "Link"} })).toBeVisible();
    expect(screen.getByRole("textbox", { name: "Start date"} })).toBeVisible();
    expect(screen.getByRole("textbox", { name: "End date"} })).toBeVisible();
    expect(screen.getByRole("button", { name: "Replace Current Banner"} })).toBeVisible();
```
This aligns more closely to how a user would interact with the page; they can't easily tell if these inputs are in a `<form>`, they can only tell whether the inputs are present on the page.

You might object that this becomes a really verbose & redundant test, and I would agree. For one, we don't have any conditional logic in `<AdminBannerForm>` for whether or not to display the form, so there isn't really a way for this test to fail. For two, the next test verifies all the inputs are present anyway, by typing in / clicking on them. In fact I _think_ `userEvent` also verifies inputs' _visibility_ before typing in them? So I would also be fine if we just delete this test.

services/ui-src/src/components/forms/AdminBannerForm.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { testA11y } from "utils/testing/commonTests";
+
+const mockWriteAdminBanner = jest.fn();
+window.HTMLElement.prototype.scrollIntoView = jest.fn();
+
+const adminBannerFormComponent = (writeAdminBanner: Function) => (
+  <RouterWrappedComponent>
+    <AdminBannerForm
+      writeAdminBanner={writeAdminBanner}
+      data-testid="test-form"
```
benmartin-coforma, line 16: And hey, if we delete that test, we won't need to pass in this prop!

```suggestion
```

services/ui-src/src/components/forms/AdminBannerForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+          {submitting ? <Spinner size="md" /> : "Replace Current Banner"}
+        </Button>
+      </Flex>
+    </>
+  );
+};
+
+interface Props {
+  writeAdminBanner: Function;
+  [key: string]: any;
```
benmartin-coforma, line 105: And if we remove that prop from the test file, we'll be able to remove it here too!

```suggestion
```

services/ui-src/src/components/pages/Admin/AdminPage.tsx
```diff
@@ -0,0 +1,174 @@
+import { useContext, MouseEventHandler } from "react";
+// components
```
benmartin-coforma, line 2:
```suggestion
```

services/ui-src/src/components/pages/Admin/AdminPage.tsx
```diff
@@ -0,0 +1,174 @@
...
+  Spinner,
+} from "@chakra-ui/react";
+import {
+  AdminBannerContext,
+  AdminBannerForm,
+  Banner,
+  ErrorAlert,
+  PageTemplate,
+} from "components";
+// utils
```
benmartin-coforma, line 19:
```suggestion
```

services/ui-src/src/components/pages/Admin/AdminPage.tsx
```diff
@@ -0,0 +1,174 @@
...
+import { convertDateUtcToEt, useStore } from "utils";
+// verbiage
```
benmartin-coforma, line 21:
```suggestion
```

services/ui-src/src/components/pages/Admin/AdminPage.tsx
```diff
@@ -0,0 +1,174 @@
...
+import verbiage from "verbiage/pages/admin";
+
+export const AdminPage = () => {
+  const { deleteAdminBanner, writeAdminBanner } =
+    useContext(AdminBannerContext);
+
+  // state management
```
benmartin-coforma, line 28:
```suggestion
```

services/ui-src/src/types/banners.ts
```diff
@@ -0,0 +1,23 @@
...
+  link?: string;
+  [key: string]: any;
+}
+
+export interface AdminBannerData extends BannerData {
+  key: string;
+  startDate: number;
+  endDate: number;
+  isActive?: boolean;
+}
```
benmartin-coforma, line 15: Naturally, if we change the type def in the API, we should change it here to match. I think this will match my previous suggestion there?

```suggestion
export interface BannerData {
  title: string;
  description: string;
  link?: string;
  key: string;
  startDate: number;
  endDate: number;
  isActive?: boolean;
}
```

services/ui-src/src/types/states.ts
```diff
@@ -1,6 +1,25 @@
 import { ParentPageTemplate, PageData, Report } from "types/report";
 import React from "react";
-import { HcbsUser } from "types";
+import { AdminBannerData, ErrorVerbiage, HcbsUser } from "types";
+
+// initial admin banner state
+export interface AdminBannerState {
+  // INITIAL STATE
```
benmartin-coforma, line 7: I understand that these comments are copied from elsewhere, but they are wrong. While an interface can be used to represent state, it cannot contain state, initial or otherwise.

The initial state is actually defined in state/useStore.ts, lines 32-57.

```suggestion
export interface AdminBannerState {
```

services/ui-src/src/utils/api/requestMethods/banner.test.ts
```diff
@@ -0,0 +1,31 @@
+import { getBanner, writeBanner, deleteBanner } from "./banner";
+// utils
```
benmartin-coforma, line 2:
```suggestion
```

services/ui-src/src/components/banners/Banner.tsx
```diff
@@ -0,0 +1,20 @@
...
+      bannerData && (
+        <Alert title={title} description={description} link={link} {...props} />
+      )
+    );
+  } else return <></>;
+};
+
+interface Props {
+  bannerData: BannerData | undefined;
+  [key: string]: any;
```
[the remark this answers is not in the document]
ailZhou: this one i highly agree with removing, why bother even having a BannerData type if we are going to be able to ignore it.

services/ui-src/src/components/forms/AdminBannerForm.test.tsx
ailZhou: i went with making sure the elements is on the page since you wrote it out for me anyway

services/app-api/types/banner.ts
benmartin-coforma, line 1: We should probably make startDate and endDate optional here, to match the UI type definition.

services/ui-src/src/components/pages/Admin/AdminPage.tsx
```diff
@@ -63,11 +59,11 @@ export const AdminPage = () => {
                     </Text>
                     <Text sx={sx.currentBannerDate}>
                       Start Date:{" "}
-                      <span>{convertDateUtcToEt(bannerData?.startDate)}</span>
+                      <span>{convertDateUtcToEt(bannerData?.startDate!)}</span>
```
benmartin-coforma, line 62: I think this needs a change, but I'm not sure what it should be.

For one, the `?` in `bannerData?.startDate` was never necessary here; we know that `bannerData` exists from the check on line 51.

More importantly, the `!` worries me. We're telling Typescript "trust me, this value will always exist"... but you and I know that it _won't_ always exist, because the field in AdminBannerForm is optional. What will we display for a banner with no defined start date? Today's date? An error? Nothing?

I haven't tested this, but I suspect (based on my reading of `convertDateUtcToEt`) that it will be today's date, which would be incorrect behavior, right? Do you think the best behavior would be to display nothing instead? Something like: `{ bannerData.startDate && <span{convertDate...` perhaps?

If you've already tested this scenario and are satisfied with the behavior, my apologies - but maybe we should make that behavior explicit? Something like: `bannerData.startDate ?? new Date().valueOf()` perhaps? Or maybe a change within `convertDateUtcToEt`, so that it accepts `number | undefined`?

And whatever change we make here, we should also make 4 lines down for `endDate`.

benmartin-coforma (commented):
This is looking really good! My remaining comments are quibbles about types.

services/ui-src/src/components/pages/Admin/AdminPage.tsx
ailZhou: I'm glad you gave feedback on this because I was thinking about this before starting work today and wasn't certain what to do. 

The start & end date field isn't suppose to be optional when filling out the admin banner form as there's a set time frame when the banner would run, allowing for scheduling ahead of time. I think this is why it was split as a separate type that inherited from a banner parent type. There's suppose to be a validation check that would make sure the user enter's the date fields in but we currently don't have validation set up yet.

Anyway, I decided to go with your suggestion of the conditional to check if the value exist and to display nothing if somehow the user managed to save without it.

benmartin-coforma (approved):
Brilliant. Thank you for all of this!

angelaco11 (approved):
Tested on cloudfront and the banner editor looks great!

PR #63, opened 2024-11-25
Add script to clean up old Playwright test results

benmartin-coforma:
### Description
It does what it says on the tin. Hopefully.

The JS script in here is a translation of the Python script used for the same purpose in MFP and MCR. It may or may not work - we'll see!

One choice I made was to rework the command-line arguments to the script. Since both the script (removeOldFolders.js) and its invocation (in delete-pages.yml) are code committed to the repo, I don't see anything inherently more difficult about changing the max folder age in the code (if it ever needs to be changed). Therefore, I may as well hard-code it. That only left the directory argument, so I changed it from a named one (`node removeOldFolders.js --folder-name="foo"`) to a positional one (`node removeOldFolders.js foo`).

### Related ticket(s)
CMDCT-4062

---
### How to test
1. Wait for a test in the `play-correct` branch to run, successfully committing its results to this branch
2. Wait for 30 more days
3. See that those (now old) test results have been deleted

### Important updates
n/a

---
### Author checklist


benmartin-coforma:
Obsolete (and wouldn't have worked). See #96

PR #67, opened 2024-12-02
3/x Create Report by Name - Add Modal to name report

ajaitasaini:
### Related ticket(s)

CMDCT-

---
### Important updates

---
### Author checklist

---

### DIFF
--- a/services/app-api/forms/qm.ts
+++ b/services/app-api/forms/qm.ts
@@ -168,24 +168,6 @@ export const qmReportTemplate: ReportTemplate = {
         measureTemplate: MeasureTemplateName.StandardMeasure,
       },
     ],
-    optionGroups: {
-      rulesOne: [
-        {
-          cmit: 888,
-          required: true,
-          stratified: false,
-          measureTemplate: MeasureTemplateName.StandardMeasure,
-        },
-      ],
-      rulesTwo: [
-        {
-          cmit: 999,
-          required: true,
-          stratified: false,
-          measureTemplate: MeasureTemplateName.StandardMeasure,
-        },
-      ],
-    },
   },
   measureTemplates: {
     [MeasureTemplateName.StandardMeasure]: {
--- a/services/app-api/handlers/reports/buildReport.ts
+++ b/services/app-api/handlers/reports/buildReport.ts
@@ -1,7 +1,12 @@
 import KSUID from "ksuid";
 import { qmReportTemplate } from "../../forms/qm";
 import { putReport } from "../../storage/reports";
-import { Report, ReportStatus, ReportType } from "../../types/reports";
+import {
+  Report,
+  ReportStatus,
+  ReportOptions,
+  ReportType,
+} from "../../types/reports";
 import { User } from "../../types/types";
 
 const reportTemplates = {
@@ -11,7 +16,7 @@ const reportTemplates = {
 export const buildReport = async (
   reportType: ReportType,
   state: string,
-  measureOptions: string[],
+  reportOptions: ReportOptions,
   user: User
 ) => {
   const report = structuredClone(reportTemplates[reportType]) as Report;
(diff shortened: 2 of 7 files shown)


services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
```diff
@@ -23,14 +25,16 @@ import arrowLeftIcon from "assets/icons/arrows/icon_arrow_left_blue.png";
...
   const [reports, setReports] = useState<Report[]>([]);
   const { intro, body } = dashboardVerbiage;
   const fullStateName = StateNames[state as keyof typeof StateNames];
 
+  // TO-DO: update accordingly - if a user is an admin or a read-only type, use the selected state, otherwise use their assigned state
+  const activeState = state;
+
   useEffect(() => {
     if (!isReportType(reportType) || !isStateAbbr(state)) {
       return;
```
ailZhou, line 40: Since reloadReports function is the same code as this chunk here, we can probably call reloadReports here so that we can cut down on duplicate code. This is more of a clean up than a necessity so i'm okay with it being a follow up on a different PR.

ailZhou (approved):
PR looks great! I have tiny request but not a blocker for merging it in. It can be a small add on for a different PR.

rocio-desantiago (approved):
Only thing to account for in a future ticket would be a 
TODO: add behavior for when a report name isn't added - maybe an error
Right now, it applies either the `last saved report name` or `{Name of form}` when you click "Start New" and leave the text field empty.

 when you

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
ajaitasaini: addressed it in this one! thanks for catching that

ajaitasaini:
> One thing to account for in a future ticket would be a TODO: add behavior for when a report name isn't added - maybe an error Right now, it applies either the `last saved report name` or `{Name of form}` when you click "Start New" and leave the text field empty.

A report name will be required for all reports! Right now we don't have a validation check so I left the option for a default name in case we don't enter one while testing :)

services/ui-src/src/components/modals/AddEditReportModal.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  test("AddEditReportModal top close button can be clicked", () => {
+    fireEvent.click(screen.getByText("Close"));
+    expect(mockCloseHandler).toHaveBeenCalledTimes(1);
+  });
+
+  test("AddEditReportModal bottom cancel button can be clicked", () => {
+    fireEvent.click(screen.getByText("Cancel"));
+    expect(mockCloseHandler).toHaveBeenCalledTimes(1);
+  });
```
benmartin-coforma, line 54: I usually prefer `userEvent` over `fireEvent` &mdash; `fireEvent` is a lower-level library that, well, fires a single event (like "click"). Whereas `userEvent` is higher-level and performs more surrounding checks (like ensuring the element is visible & clickable, and firing "mousedown" and "mouseup" along with "click"). That makes it closer to how users will actually interact with our site, and therefore a higher-quality test.

You will have to `import userEvent from "@testing-library/userEvent"` to make it work though.

```suggestion
  test("AddEditReportModal top close button can be clicked", () => {
    userEvent.click(screen.getByText("Close"));
    expect(mockCloseHandler).toHaveBeenCalledTimes(1);
  });

  test("AddEditReportModal bottom cancel button can be clicked", () => {
    userEvent.click(screen.getByText("Cancel"));
    expect(mockCloseHandler).toHaveBeenCalledTimes(1);
  });
```

benmartin-coforma (dismissed):
I've got a few nitpicks, but the structure you've built here is really good 👍

angelaco11 (approved):
Tested on cloudfront and everything is working as expected. I think in the future it would be cool to have a different default report name if the user doesn't provide a name. like a date/timestamp or something unique. I didn't see any tickets in the backlog regarding something like that so maybe I'll mention to @BearHanded if I don't forget.

anyway, that's for future musings and not for this PR! Good work

benmartin-coforma (approved):

PR #70, opened 2024-12-03
CMDCT-4136: adding measures to required measures page

angelaco11:
### Description

Screenshot 2024-12-03 at 2 37 32 PM

### Related ticket(s)
CMDCT-4136

---
### How to test
1. Log in as any user
2. Go to QM report and start a new one
3. Go to Required Measure Results tab
4. Make sure that the page tab matches the [figma](https://www.figma.com/design/B8IEoYKQ7DuQyx5nStls8F/HCBS?node-id=787-11139&t=z5vcrRQNrkO5Rhul-0)

### DIFF
--- a/services/app-api/forms/qm.ts
+++ b/services/app-api/forms/qm.ts
@@ -4,6 +4,12 @@ import {
   ElementType,
   ReportType,
   MeasureTemplateName,
+  ButtonLinkTemplate,
+  HeaderTemplate,
+  AccordionTemplate,
+  SubHeaderTemplate,
+  TextboxTemplate,
+  RadioTemplate,
 } from "../types/reports";
 
 export const qmReportTemplate: ReportTemplate = {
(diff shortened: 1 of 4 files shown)


services/app-api/forms/qm.ts
```diff
@@ -229,5 +260,212 @@ export const qmReportTemplate: ReportTemplate = {
...
+      title: "LTSS-1: Comprehensive Assessment and Update",
+      type: PageType.Measure,
+      substitutable: true,
+      sidebar: false,
+      elements: [
+        {
+          type: ElementType.ButtonLink,
+          label: "Return to Required Measures Results Dashboard",
+          to: "req-measure-result",
+        } as ButtonLinkTemplate,
```
angelaco11, line 274: I was getting some insane type errors that made me cast every single element like this or else tests would fail to run. I have no clue how to resolve this otherwise but would love some help with this since casting like this is untenable.

services/ui-src/src/types/report.ts
```diff
@@ -79,6 +79,7 @@ export interface MeasurePageTemplate extends FormPageTemplate {
   required?: boolean;
   stratified?: boolean;
   optional?: boolean;
+  substitutable?: boolean;
```
angelaco11, line 82: couldn't think of a better word for measures that have the option to be substituted. open to ideas here

services/app-api/forms/qm.ts
angelaco11: I removed all the elements but just a warning this might come up later and hopefully someone who's better at typescript has to deal with it instead of me

services/ui-src/src/types/report.ts
ajaitasaini: replaceable? exchangeable? tbh substitutable is fine :)

ajaitasaini (approved):
code looks good to me, also tested in cloudfront and works as expected 👍

ailZhou (approved):
lgtm!

PR #72, opened 2024-12-06
CMDCT-4065: Read Only View

angelaco11:
### Description
We want to block users from being able to edit fields in the QM reports. The work here was to hide the Create and submit report buttons and make all the editable fields greyed out so the user can read them but not edit. 

### Related ticket(s)
CMDCT-4065

---
### How to test
https://dbls5wcxpk535.cloudfront.net/
1. Log in as any state/admin user that can create reports
2. Walk through the steps to create a report. Just create it and fill out the fields, don't submit
3. Log out and then log back in as any readonly user (helpdeskuser@test.com is one)
5. On the list of reports, click "View" on the one you just created as the state user
6. Click through the report and make sure all the text/date/radio fields are visible, but greyed out and unable to be edited
7. On the Review & Submit tab, make sure that you do not see the "Submit QMS Report" button


services/ui-src/src/components/fields/DateField.tsx
```diff
@@ -48,6 +53,7 @@ export const DateField = (props: PageElementProps) => {
         value={displayValue}
         hint={parsedHint}
         errorMessage={errorMessage}
+        {...disabled}
```
benmartin-coforma, line 56: You're exactly right, and this is a good solution. Here's the commit when CMSDS started allowing passthrough props for this input type: https://github.com/CMSgov/design-system/commit/ef43d2cb1786625402b0120727546b51d4686464#diff-ebbd9086a31139490786c9439153de660e9e93f8e2e7e70fca0cb021f1aef6b3

My only suggestion, which you may take or leave, is to localize the hack spatially. Rather than half up on line 15 and half down here, I would stuff the whole thing right here. It will make for a nice clean diff when we are (someday) able to delete it.

```suggestion
        /*
         * Typescript hack. CmsdsDateField won't recognize
         * passthrough props until @cmsgov/design-system^9.0.0
         * TODO: Unhack, to just `disabled={props.disabled}`
         */
        {...{ disabled: props.disabled } }
```

services/ui-src/src/components/fields/DateField.tsx
```diff
@@ -8,6 +8,11 @@ import { DateTemplate } from "../../types/report";
 
 export const DateField = (props: PageElementProps) => {
   const dateTextbox = props.element as DateTemplate;
+  /**
+   * This is a hack to get the custom design system component to accept the
+   * disabled prop, if used directly in the component, it will throw a ts error
+   */
+  const disabled = { disabled: props.disabled };
```
benmartin-coforma, line 15: If you call out this workaround down below, you won't need it up here.
```suggestion
```

benmartin-coforma (approved):
This looks like it'll work! We must already be disabling other field types for admin users?

services/ui-src/src/components/fields/DateField.tsx
angelaco11: This is brilliant. I was trying to figure out a way to get it all in one place, but didn't think of this so thank you!!!

angelaco11:
> This looks like it'll work! We must already be disabling other field types for admin users?

[This](https://github.com/Enterprise-CMCS/macpro-mdct-hcbs/pull/72/files#diff-efb9cbac7a4c9ea706bb4e49889f19ebb3c52d7db414b45deda7ab4d4a900c78R60) line is actually doing the disabling of all elements that can be edited if the user is not an end user. I was thinking about doing locally it in each editable element (the three being DateField, TextField, RadioField), but passing it in like this to the ComposedElement propagates the disabled field to all the editable fields without any extra code in each individual field component. So even though it's slightly obscured, it feels way cleaner. I could always add a comment to that disabled prop to explain what I just explained above since now that i'm thinking about it, it's not all that clear. In fact, I will do that!

benmartin-coforma:
> > This looks like it'll work! We must already be disabling other field types for admin users?
> 
> passing it in like this to the ComposedElement propagates the disabled field to all the editable fields without any extra code in each individual field component.

That's beautiful.

benmartin-coforma (approved):

ailZhou (dismissed):
lgtm! tested on deploy and the reports are disabled so admin users can't click on it

benmartin-coforma (approved):

BearHanded (dismissed):

BearHanded (approved):

benmartin-coforma (dismissed):

PR #81, opened 2024-12-16
CMDCT-4185: adding edit button to dashboard table rows so that user can edit report name

angelaco11:
### Description
Pretty simple PR to allow users to edit report name from the dashboard table. I went pretty simple here since there is just one field. I did not use any useForm magic so let me know if that is more appropriate here.

### Related ticket(s)
CMDCT-4185

---
### How to test
https://d2nvrsyhsh8gv1.cloudfront.net/

Test as state user:
1. Log in as any state user
2. Navigate to report table
3. You should see an edit button next to every report. Click on the edit button and you should see the edit modal
4. it should say "Edit Quality Measure Set Report" and the button should say "Save"
5. The text box should be populated with the current name. Change the name and hit save and the row should update with the name you just changed to
6. Click on the edit button again and the name in the text field should reflect the current name
7. hit save without editing the name and it should keep the same name

Test as read only user:
1. Log in as any read only user (helpdeskuser@test.com)
2. Navigate to report table
3. You should not see any edit button column, as read only users will not have the power to edit report names

### DIFF
--- a/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
@@ -11,6 +11,15 @@ interface DashboardTableProps {
   openAddEditReportModal: Function;
 }
 
+/**
+ * Return the dashboard table column headers. If user is admin, there will be
+ * an extra empty row that is for the edit report name button column.
+ */
+const getHeadRow = (readOnlyUser?: boolean) =>
+  readOnlyUser
+    ? ["Submission name", "Last edited", "Edited by", "Status", ""]
+    : ["", "Submission name", "Last edited", "Edited by", "Status", ""];
+
 export const DashboardTable = ({
   reports,
   readOnlyUser,
@@ -23,14 +32,9 @@ export const DashboardTable = ({
   const editButtonText = isStateUser ? "Edit" : "View";
 
   const tableContent = {
-    caption: "Quality Measure Reports",
-    headRow: ["", "Submission name", "Last edited", "Edited by", "Status", ""],
+    caption: "QMR",
+    headRow: getHeadRow(readOnlyUser),
   };
-
-  // pop off the edit button heading if the user is read only
-  if (readOnlyUser) {
-    tableContent.headRow.shift();
-  }
 
   return (
     <Table content={tableContent}>
@@ -39,7 +43,7 @@ export const DashboardTable = ({
           {!readOnlyUser && (
             <Td fontWeight={"bold"}>
               <button onClick={() => openAddEditReportModal(report)}>
-                <Image src={editIcon} alt="Edit Report" />
+                <Image src={editIcon} alt="Edit Report Name" />
               </button>
             </Td>
           )}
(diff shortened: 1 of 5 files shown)


BearHanded (approved):

rocio-desantiago (approved):
I accidentally tapped the edit button thinking that's how I'd edit the title 😆 
It works per the story's acceptance criteria but I'd heavily debate that design. So interesting to have an edit icon to the left and an edit button to the right. Maybe something more explicit Edit name vs Edit report. But I'm just rambling.

angelaco11:
> I accidentally tapped the edit button thinking that's how I'd edit the title 😆 It works per the story's acceptance criteria but I'd heavily debate that design. So interesting to have an edit icon to the left and an edit button to the right. Maybe something more explicit Edit name vs Edit report. But I'm just rambling.

Yeah I actually agree with this, it is not great UX to have two edit buttons in the same row. I will bring this up with Megan after we get the logic in. I think there's a way to improve it but I'd like to leave it up to her.

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -17,13 +24,25 @@ export const DashboardTable = ({ reports }: DashboardTableProps) => {
...
   const tableContent = {
     caption: "Quality Measure Reports",
-    headRow: ["Submission name", "Last edited", "Edited by", "Status", ""],
+    headRow: ["", "Submission name", "Last edited", "Edited by", "Status", ""],
   };
 
+  // pop off the edit button heading if the user is read only
+  if (readOnlyUser) {
+    tableContent.headRow.shift();
+  }
```
benmartin-coforma, line 33: Incoming waffle alert.

On the one hand, I don't like throwing away functions' return values. So I'd prefer `slice(1)` to `shift()` here.

But then we have to turn this into an assignment instead: `tc.hr = tc.hr.slice(1)`, which starts to feel awkward. Especially because we just declared the thing a second ago.

But maybe that's just revealing the underlying awkwardness of the whole situation, that we're creating something we might need to modify. So maybe we want to wrap the whole thing in a larger construct? Maybe along these lines???
```ts
const tableContent = {
  caption: "QMR",
  headRow: getHeadRow(readOnlyUser),
};
...

/** Return the dashboard table column headers, which differ for admin users */
const getHeadRow = (readOnlyUser: bool) => readOnlyUser
  ? ["Submission name", "Last edited", "Edited by", "Status", ""]
  : ["", "Submission name", "Last edited", "Edited by", "Status", ""];
```

But that's an awful lot of redundancy to include/exclude a single empty string. And maybe it'll be worth it someday, when the column headers change more? For example, maybe admins get an extra column for locking/archiving - that's what's happened in our other apps.

But then again maybe not. And surely we can trust our successors to be able to introduce such constructs as they become necessary, rather than us preemptively introducing them in the hopes that they'll be (understood and) used (well) later.

And now both you and I have spent more time writing/reading about these 3 lines of code than anyone is ever likely to in the future. So, do what you want, and I'll approve it.

angelaco11: I like this way better than my hard to read implementation! I went ahead with your solution you provided. Thanks Ben!

rocio-desantiago (approved):
Great re-factor

BearHanded (approved):

PR #93, opened 2025-01-17
CMDCT-4243 - serverless v4 upgrade

peoplespete:
### Description
Now that QMR repo has shown the v4 serverless upgrade is not too too hard of a lift, we're doing it to all the repos.

### Related ticket(s)
CMDCT-4243

---
### How to test
Check out all the tests that ran (unit and playwright). You're also welcome to poke around in the app yourself:
https://d12xvfykgjoerg.cloudfront.net/

### Important updates
NA

---
### Author checklist

---


services/database/serverless.yml
```diff
@@ -62,6 +56,7 @@ provider:
             - "s3:GetObject"
             - "s3:ListBucket"
             - "s3:PutObject"
+          Resource: "*"
```
JonHolman, line 59: How did it work without this before? :)

services/database/serverless.yml
```diff
@@ -1,19 +1,13 @@
...
   - serverless-dynamodb
   - serverless-stack-termination-protection
-  - serverless-s3-bucket-helper
   - serverless-plugin-scripts
   - serverless-offline
 
-s3BucketHelper:
-  loggingConfiguration:
-    destinationBucketName: ${env:LOGGING_BUCKET, ssm:/configuration/${self:custom.stage}/s3/accessLogsBucket, ssm:/configuration/default/s3/accessLogsBucket}
-    logFilePrefix: ${env:LOGGING_BUCKET, ssm:/configuration/${self:custom.stage}/s3/accessLogsPrefix, ssm:/configuration/default/s3/accessLogsPrefix}
```
JonHolman, line 15: We don't need these?

services/ui-auth/serverless.yml
```diff
@@ -26,16 +26,9 @@ provider:
...
   - serverless-plugin-scripts
-  - serverless-bundle
   - serverless-iam-helper
-  - serverless-s3-bucket-helper
   - "@enterprise-cmcs/serverless-waf-plugin"
 
-s3BucketHelper:
-  loggingConfiguration:
-    path: ${env:IAM_PATH, ssm:/configuration/${self:custom.stage}/iam/path, ssm:/configuration/default/iam/path, "/"}
-    permissionsBoundary: ${env:IAM_PERMISSIONS_BOUNDARY, ssm:/configuration/${self:custom.stage}/iam/permissionsBoundaryPolicy, ssm:/configuration/default/iam/permissionsBoundaryPolicy, ""}
```
JonHolman, line 37: We don't need these applied directly to the bucket now?

services/ui-src/serverless.yml
```diff
@@ -1,19 +1,13 @@
...
 plugins:
   - serverless-plugin-scripts
   - serverless-s3-sync
   - serverless-stack-termination-protection
-  - serverless-s3-bucket-helper
   - serverless-cloudfront-invalidate
 
-s3BucketHelper:
-  loggingConfiguration:
-    destinationBucketName: ${env:LOGGING_BUCKET, ssm:/configuration/${self:custom.stage}/s3/accessLogsBucket, ssm:/configuration/default/s3/accessLogsBucket}
```
JonHolman, line 14: We don't need these directly on the bucket now?

JonHolman (dismissed):

services/database/serverless.yml
peoplespete: I think this stuff was just copied in without understanding it. The serverless-s3-bucket-helper goes through the config for s3 buckets and enforces certain properties. In many cases I found, there was no s3 bucket in the serverless file so this was actually doing nothing ever.

services/ui-auth/serverless.yml
peoplespete: There's no bucket in this file. I suspect this config was just copied in when this file was initially created.

services/database/serverless.yml
peoplespete: It broke the build so I could be wrong about some of this...

peoplespete: Ok I think I figured it out. It is if there are no functions AND no IAM roles being created, then you don't need that section.

peoplespete: Confirmed, that seems to be how it works.

JonHolman (approved):

PR #99, opened 2025-01-28
Random Clean Up

ailZhou:
### Description

- Fix header sub bar's sticky position 
- Fix footer contact us scroll to top 
- Added a Text Area component
- Added autosaving for nested children input field components 
- Hardcoded width for sidebar nav to keep close animation consistent
- Added more left padding for full size pages
- Updated children text field width to match figma's

https://github.com/user-attachments/assets/e6be390d-512c-458a-9add-c1d8095e65a7

**Note**: I am aware that switching from yes/no radio button does not clear out that text field. Will be dealt with in a more thorough look at the auto save in a different PR.

### Related ticket(s)

CMDCT-4275

---
### How to test

Deploy Link: https://d3f0cw5x2cj6vy.cloudfront.net/

2) Create a new report and click around.
3) When you scroll up and down, the light blue bar should follow the header
4) In [Required Measure Results], click the edit button for LTSS-1. The full page view should have a little more padding to the left.
5) For the question **Did you follow the [reportYear] Technical Specifications?**, select [No] and you should see a text area box instead of a small text field.
6) Fill out the form, let it auto save.
7) Return to the form and see that the radio button's child fields still have their data
8) Scroll to the footer and click contact us, when the contact us page loads, you should be at the top of the page and not the bottom.

### Important updates

N/A

---
### Author checklist

---


services/ui-src/src/components/fields/RadioField.tsx
```diff
@@ -7,11 +7,23 @@ import { parseCustomHtml } from "utils";
...
-export const formatChoices = (choices: ChoiceTemplate[], answer?: string) => {
-  return choices.map((choice) => {
-    const formatFields = choice?.checkedChildren ? (
+export const formatChoices = (
+  parentKey: string,
+  choices: ChoiceTemplate[],
+  answer?: string
+) => {
+  return choices.map((choice, choice_index) => {
+    const children = choice?.checkedChildren?.map((child, child_index) => {
```
benmartin-coforma, line 16: My nitpick here is to prefer camelCasing over snake_casing for `choiceIndex` and `childIndex`.

If you are willing to refactor the method to include fewer question marks, fewer ternaries, and 0.5 less variables, here's my attempt:

```tsx
export const formatChoices = (
  parentKey: string,
  choices: ChoiceTemplate[],
  answer?: string
) => {
  return choices.map((choice, choiceIndex) => {
    if (!choice?.checkedChildren) {
      return {
        ...choice,
        checked: choice.value === answer,
        checkedChildren: [],
      };
    }

    const children = choice.checkedChildren.map((child, childIndex) => ({
      ...child,
      formKey: `${parentKey}.value.${choiceIndex}.checkedChildren.${childIndex}`,
    }));

    const checkedChildren = [
      <Box sx={sx.children}>
        <Page elements={children} />
      </Box>,
    ];

    return {
      ...choice,
      checkedChildren,
      checked: choice.value === answer,
    };
  });
};
```

services/ui-src/src/components/fields/TextAreaField.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  return (
+    <Box>
+      <CmsdsTextField
+        id={key}
+        name={key}
+        label={labelText || ""}
+        hint={parsedHint}
+        onChange={(e) => onChangeHandler(e)}
+        onBlur={(e) => onBlurHandler(e)}
```
benmartin-coforma, line 47: Suggestion: use the point-free style here
```suggestion
        onChange={onChangeHandler}
        onBlur={onBlurHandler}
```

services/ui-src/src/components/fields/RadioField.tsx
ailZhou: added!

services/ui-src/src/components/fields/TextAreaField.tsx
ailZhou: also added!

services/ui-src/src/components/fields/TextField.tsx
```diff
@@ -43,8 +43,8 @@ export const TextField = (props: PageElementProps) => {
         name={key}
         label={labelText || ""}
         hint={parsedHint}
-        onChange={(e) => onChangeHandler(e)}
-        onBlur={(e) => onBlurHandler(e)}
+        onChange={onChangeHandler}
+        onBlur={onBlurHandler}
```
benmartin-coforma, line 47: Thank you! ❤️

BearHanded (approved):

BearHanded (dismissed):

services/ui-src/src/components/fields/TextAreaField.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+const mockedTextAreaElement = {
+  type: TextAreaField,
+  label: "test label",
+  helperText: "helper text",
+};
+
+const textAreaFieldComponent = (
+  <TextAreaField
+    element={mockedTextAreaElement as unknown as PageElement}
```
benmartin-coforma, line 40: The `as unknown as Foo` double-cast is a sledgehammer that will convince Typescript to ignore even the most egregious type issues. Therefore, when you find yourself needing to use it - even within unit tests - that's usually indicative of an egregious type issue. The unfortunate thing is that the error message often doesn't point to the real issue, and some work is required to track it down.

In this case, removing the `as unknown` from the cast here gives me "Property 'value' is missing in type [...] but required in type RadioTemplate". That's not helpful because 1) I know I'm not trying to create a RadioTemplate - I'm trying to create a TextAreaBoxTemplate. And 2) I don't intend to provide a `value`, because it's not relevant to the unit test I'm trying to write.

Changing the cast to `as TextAreaBoxTemplate` (and updating the imports appropriately) gives this much more helpful message: "Types of property 'type' are incompatible. Type (props) => JSX.Element is not comparable to type 'ElementType.TextAreaField'". And sure enough, five lines above, the declaration of `mockedTextAreaElement` has `type: TextAreaField` (a reference to the component itself) rather than `type: ElementType.TextAreaField` (a value from the ElementType enum). Fixing that satisfies TS.

The general rule at play here is that TS is happy to cast an object to a type when that object is _missing_ fields that the type normally requires, but refuses to cast when that object's fields _conflict_ with those in the type. For example:
```ts
type Foo = { str: string, num: number };
const a = { str: "hello" } as Foo; // succeeds
const b = { num: 42 } as Foo; // succeeds
const c = { str: 42 } as Foo; // fails
```

Tangentially, as a style suggestion, I like to do my casts as early and specifically as possible. In this case that means changing the declaration to this, and removing all subsequent casts.
```ts
const mockedTextAreaElement = {
  type: ElementType.TextAreaField,
  label: "test label",
  helperText: "helper text",
} as TextAreaBoxTemplate;
```

That way, if there are any complaints from TS, we'll see them right away & closer to the source of the problem - as opposed to later, when we try to use the variable, and may get less specific error messages.

benmartin-coforma (approved):
Wonderful, thank you Aileen!

services/ui-src/src/components/fields/TextAreaField.test.tsx
ailZhou: i deserve this for copying the unit test straight from the textfield.test.tsx and then renaming it. thanks for pointing it out!

PR #102, opened 2025-01-31
CMDCT-4189: Setting up frontend banner validation display in banner form

angelaco11:
### Description
This PR handles the display for frontend banner validation in the admin banner form. Uses a yup schema that is passed into react hook form to handle errors automagically.

### Related ticket(s)
CMDCT-4189

---
### How to test
https://d30xzmra0k9o0v.cloudfront.net/
1. Log in as adminuser@test.com
2. Go to My Account -> Manage Account -> Banner Editor
3. In the banner editor, type something in all the fields. One by one, delete all the text from a field and observe that a red error message pops up and says that the field you're editing is required. Note: link is not a required field so this and the following steps will not apply to link. 
4. Enter some more text in the errored out fields and notice that the error goes away
5. Refresh the page and then hit the "Replace Current Banner" button without filling in any info. Notice that the title, description, start date and end date all throw required errors. Fill in all the fields and hit Replace Current Banner and notice that it works


services/ui-src/src/components/forms/AdminBannerForm.tsx
```diff
@@ -32,9 +74,11 @@ export const AdminBannerForm = ({ writeAdminBanner, ...props }: Props) => {
     };
 
     try {
+      // validate banner data before making api call
+      await validateBannerPayload(newBannerData);
```
angelaco11, line 78: this check seems like overkill since we are not allowing users to enter bad data and then we are making this same schema validation call on the backend. I could go either way on having it in here or removing it. Mostly I feel like it couldn't hurt, but does contribute to extra duplicated code. Thoughts?

services/ui-src/src/utils/validation/bannerValidation.test.ts
```diff
@@ -0,0 +1,32 @@
...
+const invalidObject = {
+  // missing key
+  title: "this is a title",
+  description: "this is a description",
+  link: "https://www.google.com",
+  startDate: 11022933,
+  endDate: 103444405,
+};
+
+describe("Test validateBannerPayload function", () => {
```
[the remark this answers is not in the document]
angelaco11: I'm ignoring this because these files are in two different services.

services/ui-src/src/components/forms/AdminBannerForm.tsx
```diff
@@ -4,18 +4,60 @@ import { Button, Flex, Spinner } from "@chakra-ui/react";
...
 import { bannerId } from "../../constants";
 import { bannerErrors } from "verbiage/errors";
-import { convertDatetimeStringToNumber } from "utils";
+import { convertDatetimeStringToNumber, checkDateCompleteness } from "utils";
 import { ElementType, ErrorVerbiage } from "types";
+import { object, string } from "yup";
+import { yupResolver } from "@hookform/resolvers/yup";
+import { validateBannerPayload } from "utils/validation/bannerValidation";
+
+const bannerFormValidateSchema = object().shape({
```
benmartin-coforma, line 13: Do think adding a doc comment like this would be too wordy?
```tsx
/**
 * This schema is used for live validation as the user fills out the form.
 * It differs from the schema in bannerValidation.ts,
 * because the shape of the data we bind to the form
 * differs from the shape of the data we send to the backend.
 * We transform from one shape to the other in `onSubmit()`.
 */
```

services/ui-src/src/components/forms/AdminBannerForm.tsx
benmartin-coforma: I feel the same way - it's technically redundant, but I don't see it hurting us. Could take it or leave it. Does anyone else have a stronger opinion here?

services/app-api/utils/bannerValidation.ts
```diff
@@ -9,6 +9,7 @@ const bannerValidateSchema = object().shape({
   link: string().url().notRequired(),
   startDate: number().notRequired(),
   endDate: number().notRequired(),
+  isActive: boolean().notRequired(),
```
benmartin-coforma, line 12: I think I'm getting a little lost, because I'm not sure where this `isActive` flag is coming from, but it doesn't feel like a great idea to me... Whether or not a banner is active depends upon the Rest of the Universe (i.e., the current date & time), so I don't think we should be storing that flag in our database. Right?

services/ui-src/src/components/forms/AdminBannerForm.tsx
angelaco11: I love this comment. Gonna put it in but it does bloat the file. So therefore, I am moving this validation schema to bannerValidation.ts

services/app-api/utils/bannerValidation.ts
angelaco11: Ben and discussed this and agreed that isActive could easily be replaced by just checking that the date is in between the start and end date. i will probably write a ticket to address this

benmartin-coforma (approved):

rocio-desantiago (dismissed):
Manually tested and works as expected ✅ 
One thing to note in the future, might be the error message for the LINK:

> bannerLink.answer must be a valid URL

banner link with error message
Also not sure, if we want to include directions on how the https:// is required or if we want to autofill that for the user to make it more user-friendly.
banner link with error message

angelaco11:
> Manually tested and works as expected ✅ One thing to note in the future, might be the error message for the LINK:
> 
> > bannerLink.answer must be a valid URL
> 
> banner link with error message Also not sure, if we want to include directions on how the https:// is required or if we want to autofill that for the user to make it more user-friendly. banner link with error message

I updated the error message for link to correspond with design demands

rocio-desantiago (approved):
new validation messages ✅

PR #107, opened 2025-02-05
CMDCT-4302 & CMDCT-4309 - UI Bucket should not be a website in S3 and enforce SSL on Serverless Buckets

JonHolman:
### Description
CMDCT-4302 & CMDCT-4309 - UI Bucket should not be a website in S3 and enforce SSL on Serverless Buckets

### Related ticket(s)

CMDCT-4302
CMDCT-4309

---
### Important updates

---
### Author checklist

---

### DIFF
--- a/src/slsV4BucketPolicies.ts
+++ b/src/slsV4BucketPolicies.ts
@@ -1,0 +1,75 @@
+import {
+  S3Client,
+  ListBucketsCommand,
+  GetBucketPolicyCommand,
+  PutBucketPolicyCommand,
+} from "@aws-sdk/client-s3";
+
+const s3Client = new S3Client({ region: "us-east-1" });
+
+async function getServerlessBuckets(): Promise<string[]> {
+  const { Buckets } = await s3Client.send(new ListBucketsCommand({}));
+  return (Buckets || [])
+    .map((bucket) => bucket.Name!)
+    .filter((name) => name.startsWith("serverless-framework-"));
+}
+
+async function getBucketPolicy(bucketName: string): Promise<string | null> {
+  try {
+    const { Policy } = await s3Client.send(
+      new GetBucketPolicyCommand({ Bucket: bucketName })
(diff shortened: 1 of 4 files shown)


src/slsV4BucketPolicies.ts
```diff
@@ -0,0 +1,75 @@
+import {
```
peoplespete, line 1: I wonder if this is possible (not sure) to do this with serverless:
https://www.serverless.com/framework/docs/providers/aws/guide/serverless.yml#deployment-bucket
I guess the problem with that link is that it's for the stack specific deployment bucket vs the global all stacks of all branches. But let me know if you find anything worth thinking about.

src/slsV4BucketPolicies.ts
```diff
@@ -0,0 +1,75 @@
...
+  const policy = JSON.stringify({
+    Version: "2012-10-17",
+    Statement: [
+      {
+        Sid: "AllowSSLRequestsOnly",
+        Action: "s3:*",
+        Effect: "Deny",
+        Resource: [
+          `arn:aws:s3:::${bucketName}`,
+          `arn:aws:s3:::${bucketName}/*`,
```
peoplespete, line 41: Since these buckets are used by lots of different branches deploys, we probably are going to have to be looser than a specific bucket here. I'm concerned this would make it work just for the last person who's deployed.

JonHolman: I'm confused.  Can we huddle?

peoplespete: I misunderstood, looks great.

src/slsV4BucketPolicies.ts
peoplespete: There's not a way with compose at this point so ignore this.

peoplespete (approved):
👨‍🎤

PR #116, opened 2025-02-13
Spikes: Element Ids, Clear actions

BearHanded:
### Description

Apologies in advance, this has very much been an exercise in yak shaving. Set out to just get clear functionality working and kept brushing up against other issues or requirements to do so.

This PR:
- Adds Ids to report page elements. These are required fields in the schemas moving forward, and a report won't generate without them.
- Introduces a clear and reset action ("Reset" being a hard reset, clear just removing all but the whitelisted answers)
- Fixes behavior in the quality measure element and radio to save correctly, and ties into the recent fix to display at the correct time
- Removes the cmit entry in state because it was leaning on needing a setMeasure call in advance to prep loading a page. Moved that cmitId to the measure page objects themselves. That gets added during buildReport from the uid.

Gotchas:
- **I owe the reset files unit tests** (Got me)
- Children still need to be cleared in the reset file, but that's just a matter of looking up the children and adding a recursive call to the method
- The reporting button has an upcoming ticket to do the work, but currently triggers the clearing method on clicking "no". This sets the report store to the correct state, but has not been tied to the react hook form.

### Related ticket(s)

CMDCT-4241, CMDCT-4169

---
### Important updates

---
### Author checklist

---


ailZhou (approved):
There's a lot of good changes in this PR, the only thing that concerns me is the cmitId but I don't think there will be an issue with merging that in atm. This will cause merge conflicts with my pr 😭

services/app-api/utils/reportValidation.ts
angelaco11, line 1: i'm glad it was this easy to edit this schema!!!

angelaco11 (approved):

PR #117, opened 2025-02-14
QMS - Substitute Measure Functionality

ailZhou:
### Description

This PR handles the modal that pops up when you click the [**Substitute measure**] button.

https://github.com/user-attachments/assets/a887e414-5b16-4c06-8d19-d135d27cf0d2

Notable Changes:
1. The buttonLinkElement had to be altered to be able to guess when a measure has been swapped so that the return  would go to the right page. It will always look for a path we input but if we leave it empty, it will assume it's a measure page and try to go back to the appropriate results page.
2. The URL has been altered to not have the CMIT number appended to the id when the reports get generated. Since all measures are unique, that didn't seem needed anymore.

### Related ticket(s)

CMDCT-4138

---
### How to test

2) Generate a report
3) Enter the report and select [Required Measure Results] on the sidebar menu.
4) Click any [Substitute measure] on either LTSS-1 or LTSS-2. 
5) The modal will popup, select the [Yes] radio button and then click [Substitute Measure]
6) The measure will now be replaced with the substituted measure and the measure should open correctly.

### Important updates

---
### Author checklist

---


services/ui-src/src/components/report/MeasureTable.tsx
```diff
@@ -35,11 +35,28 @@ export const MeasureTableElement = (props: PageElementProps) => {
...
+    if (selectMeasure) {
+      const substitute = selectMeasure.substitutable?.toString();
+      const measure = measures.find((measure) =>
+        measure.id.includes(substitute!)
+      );
+      if (report && measure) {
+        measure.required = true;
+        selectMeasure.required = false;
+
+        await putReport(report);
```
BearHanded, line 48: I think this should all be wrapped into a store action, and let that handle this logic and putReport()

ailZhou: swapped over to using the store actions

services/app-api/utils/reportValidation.ts
```diff
@@ -128,7 +128,7 @@ const radioTemplateSchema = object().shape({
 const buttonLinkTemplateSchema = object().shape({
   type: string().required(ElementType.ButtonLink),
   label: string().required(),
-  to: string().required(),
+  to: string().optional(),
```
rocio-desantiago, line 131: are `.notRequired()` and `.optional()` the same? Just wondering since I've seen the pattern for `.notRequired()` being used if `.optional()` has a difference.

rocio-desantiago (commented):
Just a question. Gonna test in the morning 😄

services/app-api/utils/reportValidation.ts
ailZhou: That's a really good question, i didn't even think about it 🤔 . 

@BearHanded & @angelaco11 what is the difference?

services/app-api/utils/tests/mockReport.ts
```diff
@@ -141,7 +141,7 @@ export const invalidRadioCheckedChildrenReport = {
       id: "LTSS-1",
       title: "LTSS-1: Comprehensive Assessment and Update",
       type: PageType.Measure,
-      substitutable: true,
+      substitutable: MeasureTemplateName["FASI-2"],
```
ajaitasaini, line 144: nonblocking comment but isn't LTSS-1 only substitutable with FASI-1, and not FASI-2?

ailZhou: oh it's just a mock report for unit testing but i can make it more data accurate if you want

ajaitasaini: i don't want to block you so it's up to you, worst case we just update it later

BearHanded (approved):

PR #129, opened 2025-02-27
QM Table: State vs Form fix

BearHanded:
### Description

React Hook Form and our store are enemies some days, but today we make them friends.
The QM component needs to update when radio buttons are clicked and be informed.

React Hook form:
- Hands off its updates on change successfully but sometimes has an empty initial state for useWatch()
- Our store always has the correct initial state, but only updates on blur actions. We need the info on a click.

This custom hook merges those two together, given an elementId on the current page. It is a little chattier than normal, and doesn't make sense everywhere, but it does make sense for QM tables.

### Related ticket(s)

N/A

---
### How to test

https://d1pa8viu70th2x.cloudfront.net/
- Login as state user
- Go to a report measure page
- Fill out reporting type, it should live update
- Refresh, the Edit link should still be highlighted
- Clear the measure, it should also clear

### Important updates

---
### Author checklist

- [ x I have performed a self-review of my code
---

### DIFF
--- a/services/ui-src/src/utils/state/hooks/useLiveElement.ts
+++ b/services/ui-src/src/utils/state/hooks/useLiveElement.ts
@@ -3,6 +3,16 @@ import { useWatch } from "react-hook-form";
 import { PageElement } from "types";
 import { useStore } from "utils";
 
+/**
+ * This hook gives the most up to date version of an element on the current page.
+ * This solves two problems:
+ *   - The react-hook-form hooks can init empty but they are always the most up to date after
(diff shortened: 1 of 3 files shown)


services/ui-src/src/utils/state/hooks/useLiveElement.ts
angelaco11, line 1: i'm crying at this beautiful marriage of the store and form data. may it be long and prosperous

angelaco11 (approved):
This is very good stuff. Tested and everything is behaving normally.

PR #132, opened 2025-02-28
Upgrade CMSDS to v12 and Chakra to v2

ailZhou:
### Description

This PR will bring a couple of packages to newer versions:  
- **CMS Design System**:  v3.8.0 to v12.0.0
- **Chakra**: v1.8.9 to v2.10.6
- **Jest**: v28.0.0 to v29.7.0
- **React** & **React-dom**:  v17.0.1 to v18.0.0

Important:
1. CMS Design System changed what a dropdown is. They aren't JSX.elements but ReactElements so I had to change the way it was being called. I am confused by this choice but it's not too different from what we are doing now.
2. CSSObject type is gone, it is now SystemStyleObject for Chakra.
3. Unit test is stricter with requesting you do act calls when a function triggers a useEffect. Not calling it correctly means values won't update and test will fail. If anyone knows how to call it better, I would appreciate it.

There's a couple of lingering bugs from the update, the most annoying is the sidebar navigation returning inconsistent size checks. I will deal with that when I do a sweep of the UI in a different PR.

### Related ticket(s)

CMDCT-4359

---
### How to test

Deploy Link: https://d9umoin6ntusq.cloudfront.net/

1) Sign into HCBS, any user
2) Generate a QMS report and check around
3) Check that the functionality of the forms are working and nothing looks too out of the place. Ignore any minor design imperfections.

### Important updates

---
### Author checklist

---


services/ui-src/src/components/accordions/AccordionItem.tsx
```diff
@@ -35,9 +34,9 @@ export const AccordionItem = ({ label, children, sx: sxOverride }: Props) => {
 };
 
 interface Props {
-  children?: ReactChild | ReactChild[];
+  children?: ReactNode | ReactNode[];
   label?: string;
-  sx?: CSSObject;
+  sx?: Object;
```
angelaco11, line 39: any reason why this one is just regular `Object` and not `SystemStyleObject`?

angelaco11 (approved):
Clicked around the cloudfront link and everything seems to be working as expected! good work

rocio-desantiago (approved):
Tested deploy link and all is working well.

PR #133, opened 2025-03-03
CMDCT-4358 - cdk transition sls side DO NOT MERGE

peoplespete:
# DO NOT MERGE EVER

### Description
This is the changes we need to make to the serverless setup before we convert to cdk.

### Related ticket(s)
CMDCT-4358

---
### How to test
You can double check that we haven't left out any resources that could or should be retained. At a minimum we need Cognito Userpool, all Dynamo Tables, and Cloudfront Distribution.
The playwright tests don't pass because this change breaks the app. That's fine when we are in the moment of transition but that's one reason we should never merge this :)

### Important updates
NA

---
### Author checklist

---

### DIFF
--- a/src/run.ts
+++ b/src/run.ts
@@ -1,23 +1,20 @@
 /* eslint-disable multiline-comment-style */
+/* eslint-disable no-console */
 import yargs from "yargs";
 import * as dotenv from "dotenv";
 import LabeledProcessRunner from "./runner.js";
 import { ServerlessStageDestroyer } from "@stratiformdigital/serverless-stage-destroyer";
 import { execSync } from "child_process";
 import { addSlsBucketPolicies } from "./slsV4BucketPolicies.js";
+import {
+  getAllStacksForStage,
+  getCloudFormationTemplatesForStage,
+} from "./getCloudFormationTemplateForStage.js";
 
 // load .env
 dotenv.config();
 
-const deployedServices = [
-  // "database",
-  // "uploads",
-  // "topics",
-  "app-api",
-  "ui",
-  "ui-auth",
-  "ui-src",
-];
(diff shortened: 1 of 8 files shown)


src/run.ts
```diff
@@ -1,23 +1,20 @@
...
-const deployedServices = [
-  // "database",
-  // "uploads",
-  // "topics",
-  "app-api",
-  "ui",
-  "ui-auth",
-  "ui-src",
-];
+const deployedServices = ["database", "app-api", "ui", "ui-auth", "ui-src"];
```
peoplespete, line 17: I suspect database should have been in this list already.

services/app-api/serverless.yml
```diff
@@ -212,6 +212,9 @@ resources:
           gatewayresponse.header.Access-Control-Allow-Headers: "'*'"
         ResponseType: DEFAULT_5XX
         RestApiId: !Ref ApiGatewayRestApi
+    ApiGatewayLogGroup:
```
peoplespete, line 215: Remove this per convo with Jon

peoplespete: Done.

src/serverless-stage-destroyer.ts
```diff
@@ -0,0 +1,315 @@
...
+      stacks = stacks.filter((i) =>
+        i.Tags?.find((j) => j.Key == matchTag.Key && j.Value == matchTag.Value)
+      );
+    }
+    return stacks;
+  }
+
+  private checkForProtectedStage(stage: string) {
+    // Another safeguard against destroying protected stages
+    console.log(`Not protecting prod for stage name: ${stage}`);
```
peoplespete, line 148: This is the only reason we are bringing in this library's file. We have to override the check otherwise we can't do the conversion in prod.

PR #137, opened 2025-03-07
CDMCT-4239: Hiding all other elements when user selects "CMS is reporting measure on state's behalf"

angelaco11:
### Description
This particular problem was difficult to solve because of the nature of how we use react hook form and also the way we are trying to control sibling level form values with another sibling. I decided to do a little bit of hacking in the ReportPageWrapper so this might not be ideal, but this is the best and most smooth solution I've come up with thus far. 

The solution is as follows:
1. added a boolean called `conditionallyHide` (i'm open to other names) to sibling elements that should be hidden when selected "CMS is reporting measure"
2. In ReportPageWrapper, I then listen for any form value changes. If the form has changed, I find the element that has a "measure-reporting-radio" id (id of all the reporting radios) and i check the answer. If it's "no" (no, CMS is reporting), i set a state variable called `hideElements` (open to new names for that too) to true. if "yes" i set to false
3. Then, I pass that `hideElements` state to the Page component. Then in the Page component, I will not render any element that has the `conditionallyHide` boolean && `hideElements` is set to true based on the `measure-reporting-radio` selection. 

One vaguely unrelated thing I did in this PR was to remove the "Is the state reporting on this measure?" for all pages except for LTSS-6, LTSS-7 and LTSS-8 per Cam's request [here](https://stratiformworkspace.slack.com/archives/C06V6EKEL74/p1741268828891719)

### Related ticket(s)
CMDCT-4239

---
### How to test
https://d3r1jw3z8zgool.cloudfront.net/
2. Fill out the form. Notice that as you toggle between "yes" and "no" on the "Is the state reporting on this measure?" question, "no" will hide all the elements and "yes" will bring them back. Test out various configurations of the form, like fill out a bunch of stuff when "yes" is selected, toggle to "no" so the elements all go away, and then toggle "yes" to bring back and make sure all the questions come back with nothing filled out.

### DIFF
--- a/services/ui-src/src/components/report/ReportPageWrapper.tsx
+++ b/services/ui-src/src/components/report/ReportPageWrapper.tsx
@@ -25,11 +25,27 @@ export const ReportPageWrapper = () => {
   } = useStore();
   const { reportType, state, reportId, pageId } = useParams();
   const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [hideElements, setHideElements] = useState<boolean>(false);
   const methods = useForm({
     defaultValues: {},
     shouldUnregister: true,
     resolver: yupResolver(elementsValidateSchema),
   });
+
+  useEffect(() => {
+    const formValues = methods.getValues() as any;
+    if (Object.keys(formValues).length === 0) {
+      return;
+    }
+    const reportingRadio = formValues?.elements?.find((element: any) => {
+      return element?.id === "measure-reporting-radio";
+    });
+    if (reportingRadio?.answer === "no") {
+      setHideElements(true);
+    } else {
+      setHideElements(false);
+    }
+  }, [methods.getValues()]);
 
   const navigate = useNavigate();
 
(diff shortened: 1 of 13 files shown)


services/ui-src/src/components/report/ReportPageWrapper.tsx
```diff
@@ -25,12 +25,28 @@ export const ReportPageWrapper = () => {
...
     resolver: yupResolver(elementsValidateSchema),
   });
 
+  useEffect(() => {
+    const formValues = methods.getValues() as any;
+    if (Object.keys(formValues).length === 0) {
+      return;
+    }
+    const reportingRadio = formValues?.elements?.find((element: any) => {
+      return element?.id === "measure-reporting-radio";
```
BearHanded, line 41: I still am very hesitant about this component caring about specific ids. Is there instead a way to make the element care about the id it is controlled by instead?

This report wrapper will be used for future reports and adding special cases to it will confuse future development.

angelaco11:
New PR for this here https://github.com/Enterprise-CMCS/macpro-mdct-hcbs/pull/147

PR #138, opened 2025-03-07
CMDCT-4239: a second attempt to hide elements

angelaco11:
### Description
This is a second and simpler and maybe better version of https://github.com/Enterprise-CMCS/macpro-mdct-hcbs/pull/137


angelaco11:
closing this because my branch name was making serverless deploy fail

PR #139, opened 2025-03-07
CMDCT-4239: 2nd version

angelaco11:
### Description
This is a second and simpler and maybe better version of https://github.com/Enterprise-CMCS/macpro-mdct-hcbs/pull/137


services/ui-src/src/components/fields/DateField.tsx
```diff
@@ -23,6 +24,21 @@ export const DateField = (props: PageElementProps) => {
     setDisplayValue(dateTextbox.answer ?? "");
   }, [dateTextbox.answer]);
 
+  useEffect(() => {
+    const formValues = form.getValues() as any;
+    if (formValues && Object.keys(formValues).length === 0) {
+      return;
+    }
+    const reportingRadio = formValues?.elements?.find((element: any) => {
+      return element?.id === "measure-reporting-radio";
```
BearHanded, line 33: this still feels very hard coded,  and like it might confuse future reports. Is there any way to abstract this away from an individual id on a generic question type?

benmartin-coforma: I agree. I think our best bet is to put these ID correlations inside the report template. Maybe something like this?

```ts
{
  type: ElementType.ReportingRadio,
  label: "Is the state reporting on this measure?",
  id: "measure-reporting-radio",
  value: [
    { label: "Yes, we're reporting", value: "yes" },
    {
      label: "No, CMS is reporting",
      value: "no",
      hidesElements: [
        "measure-audited-radio",
        "measure-tech-specs-radio",
        "etc",
      ]
    },
  ],
}
```

Or maybe each hidden element could point back to the reporting radio?
```ts
{
  type: ElementType.Radio,
  label: "Is the state reporting on this measure?",
  id: "measure-audited-radio",
  hiddenBy: {
    controllerId: "measure-reporting-radio",
    controllerValue: "no"
  },
  value: [
    { label: "Yes", value: "yes" },
    { label: "No", value: "no" },
  ],
}
```

Or maybe a 3rd option?

But with any of these, the main goal being to keep the logic within the page & input components more generic.

angelaco11:
New PR for this here https://github.com/Enterprise-CMCS/macpro-mdct-hcbs/pull/147

PR #147, opened 2025-03-11
CMDCT-4239: attempt 3 to hide elements based on reporting radio selection

angelaco11:
### Description
This particular problem was difficult to solve because of the nature of how we use react hook form and also the way we are trying to control sibling level form values with another sibling. I decided to do a little bit of hacking in the ReportPageWrapper so this might not be ideal, but this is the best and most smooth solution I've come up with thus far. 

The solution is as follows:
1. added an object called `hide` (i'm open to other names) to sibling elements that should be hidden when selected "CMS is reporting measure". hide has a target id (the id of the reporting radio that it's being controlled by) and the value at which it should be hidden ("no")
2. In each sibling element (RadioField, TextField and TextAreaField), I then listen for any form value changes. If the form has changed, I find the element that has the target id from the hide object and i check the answer. If it matches the value in the hide object, i set a state variable called `hideElement` (open to new names for that too) to true. 
3. Then, I just simply do not render the element if hideElement is true

One vaguely unrelated thing I did in this PR was to remove the "Is the state reporting on this measure?" for all pages except for LTSS-6, LTSS-7 and LTSS-8 per Cam's request [here](https://stratiformworkspace.slack.com/archives/C06V6EKEL74/p1741268828891719)

### Related ticket(s)
CMDCT-4239

---
### How to test
https://d3r1jw3z8zgool.cloudfront.net/
2. Fill out the form. Notice that as you toggle between "yes" and "no" on the "Is the state reporting on this measure?" question, "no" will hide all the elements and "yes" will bring them back. Test out various configurations of the form, like fill out a bunch of stuff when "yes" is selected, toggle to "no" so the elements all go away, and then toggle "yes" to bring back and make sure all the questions come back with nothing filled out.


benmartin-coforma (approved):

services/ui-src/src/components/fields/TextAreaField.test.tsx
```diff
@@ -72,6 +76,40 @@ describe("<TextAreaField />", () => {
...
+            answer: "yes",
+            type: "reportingRadio",
+            label: "Should we hide the other radios on this page?",
+            id: "reporting-radio",
+          },
+        ],
+      });
+      render(textAreaFieldComponent);
+      const textField = screen.queryByLabelText("test label");
+      expect(textField).toBe(null);
```
benmartin-coforma, line 93: Suggestion: You're already using a more semantic assertion (`toBeVisible`) for the opposite case below; invert it here to be equally semantic.
```suggestion
      expect(textField).not.toBeVisible();
```

services/ui-src/src/components/fields/TextAreaField.test.tsx
```diff
@@ -72,6 +76,40 @@ describe("<TextAreaField />", () => {
...
+          {
+            answer: "idk",
+            type: "reportingRadio",
+            label: "Should we hide the other radios on this page?",
+            id: "reporting-radio",
+          },
+        ],
+      });
+      render(textAreaFieldComponent);
+      const textField = screen.queryByLabelText("test label");
```
benmartin-coforma, line 108: As you've seen, `queryByLabelText` will happily return `null` if the element cannot be found. In that case, the only error message Testing Library will be able to give us will be from the next line. Something like "Expected `null` to be visible, but it's not, because it's not even an element"

To get a more helpful error message out of TL, explaining what it failed to find and printing the markup where it failed to find it, we can make the tests slightly asymmetric: continue to use `queryByLabelText` in the hiding test, but use `getByLabelText` in this one.

```suggestion
      const textField = screen.getByLabelText("test label");
```

services/ui-src/src/components/fields/TextField.test.tsx
```diff
@@ -68,6 +72,40 @@ describe("<TextField />", () => {
...
+            answer: "yes",
+            type: "reportingRadio",
+            label: "Should we hide the other radios on this page?",
+            id: "reporting-radio",
+          },
+        ],
+      });
+      render(textFieldComponent);
+      const textField = screen.queryByLabelText("test label");
+      expect(textField).toBe(null);
```
benmartin-coforma, line 89: My suggestion above applies here too:
```suggestion
      expect(textField).not.toBeVisible();
```

services/ui-src/src/components/fields/TextField.test.tsx
```diff
@@ -68,6 +72,40 @@ describe("<TextField />", () => {
...
+          {
+            answer: "idk",
+            type: "reportingRadio",
+            label: "Should we hide the other radios on this page?",
+            id: "reporting-radio",
+          },
+        ],
+      });
+      render(textFieldComponent);
+      const textField = screen.queryByLabelText("test label");
```
benmartin-coforma, line 104: My suggestion above applies here too:
```suggestion
      const textField = screen.getByLabelText("test label");
```

services/ui-src/src/components/fields/RadioField.test.tsx
```diff
@@ -98,3 +102,37 @@ describe("<RadioField />", () => {
...
+          answer: "yes",
+          type: "reportingRadio",
+          label: "Should we hide the other radios on this page?",
+          id: "reporting-radio",
+        },
+      ],
+    });
+    render(RadioFieldComponent);
+    const radioField = screen.queryByText("Choice 1");
+    expect(radioField).toBe(null);
```
benmartin-coforma, line 120: I'm making all of these suggestions individually so that you can apply them with 1 click from the Github UI, if you like.
```suggestion
    expect(radioField).not.toBeVisible();
```

services/ui-src/src/components/fields/RadioField.test.tsx
```diff
@@ -98,3 +102,37 @@ describe("<RadioField />", () => {
...
+        {
+          answer: "idk",
+          type: "reportingRadio",
+          label: "Should we hide the other radios on this page?",
+          id: "reporting-radio",
+        },
+      ],
+    });
+    render(RadioFieldComponent);
+    const radioField = screen.queryByText("Choice 1");
```
benmartin-coforma, line 135: Well I guess it'll be 6 clicks. Sorry for the noise. Last one:
```suggestion
    const radioField = screen.getByText("Choice 1");
```

benmartin-coforma (commented):
You have my approval for the operational code, as before. And frankly I'm fine with the test code. My suggestions are pretty nitpicky tbh.

services/ui-src/src/components/fields/TextAreaField.test.tsx
angelaco11: i would love to do this, but if the element doesn't exist on the page, it refuses to let me call .not.toBeVisible and the error is this:
    
```
    expect(received).not.toBeVisible()

    received value must be an HTMLElement or an SVGElement.
    Received has value: null
```

Do you know how to get around that? I assume it's because I'm querying (`queryByLabelText` instead of `getByLabelText`) so the query returns null if the element doesn't exist. but if do getByLabelText, it just can't find the element and fails with this error:

`TestingLibraryElementError: Unable to find a label with the text of: test label`

benmartin-coforma: 🤦 You are totally right, that is my mistake.

angelaco11: Ok I updated to use .not.toBeInTheDocument. Thank you for this feedback!

benmartin-coforma (approved):

BearHanded (approved):

PR #157, opened 2025-03-19
Store sync: Round 2

BearHanded:
### Description

Second attempt at running this through.
- Adds a selector for getting the current page, removes the repeated querying against selected pages
- Splits the save report action to its own action. This allows the synchronous part of the ui changes to run through just fine, then handle any api requests and errors. This unblocks any rendering nonsense where an action overwrites the state of the report after the api call goes through. If we do that, it should be deliberate.
- Threads errors from api calls into the state. There will be a future story to deal with the rendering, I have tested it local by rendering the error message into the action bar as a proof of concept, but no real behavior for now.
- *Disables UI Validation that is currently broken in main*. It is preventing autosave.
- *Disables a unit test in ReportPageWrapper supporting that UI validation*

### Related ticket(s)

N/A
---
### How to test

https://d11b41cs5p226h.cloudfront.net/
Measure Footer, and Measure page keep working, but the code looks less painful.
Save/Substitute/Clear/Reset all keep working (but now they add errors to the state that we'll eventually use).

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/fields/ReportingRadioField.tsx
+++ b/services/ui-src/src/components/fields/ReportingRadioField.tsx
@@ -10,7 +10,7 @@ import { ChoiceProps } from "@cmsgov/design-system/dist/react-components/types/C
 
 export const ReportingRadioField = (props: PageElementProps) => {
   const radio = props.element as ReportingRadioTemplate;
-  const { clearMeasure, currentPageId } = useStore();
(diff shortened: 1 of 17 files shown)


ailZhou (approved):
tested on deploy, did not run into any autosaving issues this time

angelaco11 (approved):
i like this refactor. also tested and things look as expected

PR #162, opened 2025-03-21
[NO TICKET]: fixing issue with hide elements functionality 

angelaco11:
### Description
Quick PR to fix bug in hide elements functionality

I also fixed radio button error display. Errors were not bubbling up. This is currently untestable because validation has been turned off so no need to test.

### How to test
https://d2jibfd6np8f8d.cloudfront.net/
2. Fill out the form. Notice that as you toggle between "yes" and "no" on the "Is the state reporting on this measure?" question, "no" will hide all the elements and "yes" will bring them back
3. Test out various configurations of the form, like fill out a bunch of stuff when "yes" is selected, toggle to "no" so the elements all go away, and then toggle "yes" to bring back and make sure all the questions come back with nothing filled out.

### DIFF
--- a/services/ui-src/src/components/fields/RadioField.tsx
+++ b/services/ui-src/src/components/fields/RadioField.tsx
@@ -1,7 +1,7 @@
 import React, { useState, useEffect } from "react";
 import { Box } from "@chakra-ui/react";
 import { PageElementProps } from "components/report/Elements";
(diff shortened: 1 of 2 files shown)


benmartin-coforma (approved):
😔  sigh... ok

BearHanded (approved):

PR #165, opened 2025-03-24
React Hook Form Error Lifecycle

BearHanded:
### Description

With validation enabled, react hook form was requiring the entire page to be filled out before saving any answers. This PR addreses those lifecycle behaviors in a few ways. The end result is that you should be able to interact with the report on the first page, or a required measure page, and fill out partial info. Putting one answer in error should still allow you to save.

What changed?
- An error handler was added in ReportPageWrapper, that just takes those same results, and any errors, and passes them off to the normal setAnswers, before calling the save.
  - If the error array is provided to this function, it will remove any answers with error messages from the store update, and save.
  - The logic here is somewhat similar to the mergeAnswers, but a little more nuanced. We look to see what levels of an object have an error message before removing that related answer, Pass everything else through unchanged.
- The fields were also immediately tying into validation behaviors on load, eased off on type validation until changes start being made.

Note the missing lines in coverage would require me to write a function to call the untested lines, by setting up a custom error trigger. This is just testing the functionality of an external package. Purposefully skipping.

### Related ticket(s)

n/a

---
### How to test

https://d12cxufdffox8f.cloudfront.net/
- Create a report
- Contact page:
  - Autosave should trigger when only filling out the first entry
  - Putting the second entry in a bad state should still allow you to edit and save the first field.
  - Editing the second and leaving it empty should still allow you to edit the first field
- LTSS-1:
  - Yes I am reporting, and blur. The child field needs info, this is expected.
  - Refreshing the page should preserve yes, this was not previously preserved.
  - You should now be able to continue editing other fields on the page while the child field is blank. This was broken.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/fields/DropdownField.tsx
+++ b/services/ui-src/src/components/fields/DropdownField.tsx
@@ -10,6 +10,11 @@ import {
 import { parseCustomHtml } from "utils";
 
 export const DropdownField = (props: PageElementProps) => {
+  /**
+   * TODO:
+   * https://design.cms.gov/components/dropdown/?theme=core
+   * WCAG 2.1 Level AA Conformance missing - Roll as custom dropdown
+   */
   const dropdown = props.element as DropdownTemplate;
   const defaultValue = dropdown.answer ?? dropdown.options[0].value;
   const [displayValue, setDisplayValue] = useState<string>(defaultValue);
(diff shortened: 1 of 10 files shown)


rocio-desantiago (approved):
Tested and works as described in PR

angelaco11 (approved):

PR #171, opened 2025-03-27
Rate Calculation C

ailZhou:
### Description

This is the missing rate from my [earlier PR](https://github.com/Enterprise-CMCS/macpro-mdct-hcbs/pull/140). 

It's used in LTSS-6 and works the same as the other rate calculations, it's just that the data structure of the saved data is a little more complicated than the others.

### Related ticket(s)

CMDCT-4289

---
### How to test

Deploy Link: https://d1zlq4kta6e8bj.cloudfront.net/

2) Create a new QMS report
3) Click **[Edit]** on the newly created report 
4) In the sidebar menu, select to **[Required Measure Results]**
5) Look for measure **LTSS-6** and press [Edit]
6) Scroll down to the question "Which delivery systems were used to report this measure?" and select [.] Fee-For-Service.
7) In the Quality Measures Table, click the newly accessibly Edit button.
8) On the new page, you should see a Performance Rate that looks similar to this [figma design](https://www.figma.com/design/B8IEoYKQ7DuQyx5nStls8F/HCBS?node-id=2438-25091&t=FPRfE5ApBhgNMNCb-4) 
9) Enter numbers and make sure that the data is auto generating and auto calculating when they are suppose to.
 - **Important**: All the rates are multiplied by 1000 for this rate calc. So if the rate is 100%, it'll be 100000%

### Important updates

---
### Author checklist

---


services/ui-src/src/components/rates/types/NDRFields.tsx
```diff
@@ -12,24 +13,28 @@ export const NDRFields = (
...
-  const initialValues =
+
+  const defaultRates: RateSetData[] =
     assessments?.map((assess) => {
       return {
-        numerator: "",
-        denominator: "",
-        rate: "",
-        performanceTarget: "",
+        label: assess.label,
```
rocio-desantiago, line 20: Trying to figure out where we'd add the title "Short Term", "Medium Term", "Long Term" for each section as shown in the Figma.
Figma image

services/ui-src/src/types/report.ts
```diff
@@ -306,6 +306,22 @@ export type PerformanceData = {
   denominator?: number;
 };
 
+export type RateType = {
+  label: string;
```
rocio-desantiago, line 310: Not sure if we'd add `sectionTitle` as a prop 🤔 here?

rocio-desantiago (changes requested):
Could we take care of the section titles? Everything else looks good

services/ui-src/src/components/rates/types/NDRFields.tsx
ailZhou: so i added the section title by pulling it from the field label key. it's doesn't exactly say `Short Term` instead it's `Short Term Stay` but I think for now that should be fine.

rocio-desantiago (approved):
LGTM 😄

BearHanded (approved):

PR #177, opened 2025-04-03
[CMDCT-4430] Admin Dashboard buttons with no styling

BearHanded:
### Description

Adds tracking for submission count, and archival, renders them, but does not fix styles. Styles will come before the ticket finishes but this unblocks the other 2 tickets.

Centralizes the logic from roles in the table to the defined behavior by column.

View per role - 
State:
image

Help Desk:
image

Admin: 
image

### Related ticket(s)

CMDCT-4430

---
### How to test

https://d2ufbmtclylfk.cloudfront.net/
Log in as state user
Create report
Submit a report (note the ui for that just leaves you sitting idly)
Log in as admin
Dashboard should display the submission count as well as new buttons that don't work!

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
@@ -7,7 +7,7 @@ import editIcon from "assets/icons/edit/icon_edit_square_gray.svg";
 
 interface DashboardTableProps {
   reports: Report[];
-  readOnlyUser?: boolean;
+  isAdmin?: boolean;
   openAddEditReportModal: Function;
 }
 
@@ -15,14 +15,14 @@ interface DashboardTableProps {
  * Return the dashboard table column headers. If user is admin, there will be
  * an extra empty row that is for the edit report name button column.
  */
-const getHeadRow = (readOnlyUser?: boolean) =>
-  readOnlyUser
-    ? ["Submission name", "Last edited", "Edited by", "Status", ""]
+const getHeadRow = (isAdmin?: boolean) =>
+  isAdmin
+    ? ["Submission name", "Last edited", "Edited by", "Status", "#", ""]
     : ["", "Submission name", "Last edited", "Edited by", "Status", ""];
 
 export const DashboardTable = ({
   reports,
-  readOnlyUser,
+  isAdmin,
   openAddEditReportModal,
 }: DashboardTableProps) => {
   const navigate = useNavigate();
@@ -33,14 +33,14 @@ export const DashboardTable = ({
 
   const tableContent = {
     caption: "Quality Measure Reports",
-    headRow: getHeadRow(readOnlyUser),
+    headRow: getHeadRow(isAdmin),
   };
 
   return (
     <Table content={tableContent}>
       {reports.map((report) => (
         <Tr key={report.id}>
-          {!readOnlyUser && (
+          {!isAdmin && (
             <Td fontWeight={"bold"}>
               <button onClick={() => openAddEditReportModal(report)}>
                 <Image src={editIcon} alt="Edit Report Name" />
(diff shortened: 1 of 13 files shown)


services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -33,14 +33,14 @@ export const DashboardTable = ({
...
-    headRow: getHeadRow(readOnlyUser),
+    headRow: getHeadRow(isAdmin),
   };
 
   return (
     <Table content={tableContent}>
       {reports.map((report) => (
         <Tr key={report.id}>
-          {!readOnlyUser && (
+          {!isAdmin && (
```
angelaco11, line 43: hmmm......wouldn't this be isAdmin now? since it was previously readOnlyUser (not admin). The edit button should only be available for state and admin users and hidden for read only users. But I logged in as helpdeskuser which is a read only and am seeing the edit button:

Screenshot 2025-04-03 at 12 43 56 PM

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -7,22 +7,22 @@ import editIcon from "assets/icons/edit/icon_edit_square_gray.svg";
...
 
 /**
  * Return the dashboard table column headers. If user is admin, there will be
  * an extra empty row that is for the edit report name button column.
  */
-const getHeadRow = (readOnlyUser?: boolean) =>
-  readOnlyUser
-    ? ["Submission name", "Last edited", "Edited by", "Status", ""]
+const getHeadRow = (isAdmin?: boolean) =>
+  isAdmin
```
angelaco11, line 19: Same as [here](https://github.com/Enterprise-CMCS/macpro-mdct-hcbs/pull/177/files#r2027399207), shouldn't this be !isAdmin? This logic sucks (i wrote it) but the "" in the second array is the column title for edit buttons that should only show up for state and admin users. does that make sense?

angelaco11: Actually........I think this still needs a read only check instead of an admin check. since state users are not admin correct??? but they should still have access to the edit button. so therefore should only NOT show the button for read only and show it for all other users. correct me if i'm wrong

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
angelaco11: but actually like [here](https://github.com/Enterprise-CMCS/macpro-mdct-hcbs/pull/177#discussion_r2027401487), i'm pretty sure this needs to be a read only check again. state users need to see this button but they aren't admin

BearHanded: Updated, it should be more explicit now

ailZhou (approved):
tested on deploy, submission count went up after i submitted the report

ailZhou (approved):

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -1,46 +1,45 @@
...
   const navigate = useNavigate();
-  const store = useStore();
-  const user = store.user;
-  const isStateUser = user?.userRole === UserRoles.STATE_USER;
-  const editButtonText = isStateUser ? "Edit" : "View";
+  const { userIsAdmin, userIsEndUser } = useStore().user ?? {};
+
+  // Translate role to defined behaviors
+  const showEditNameColumn = userIsEndUser;
+  const showReportSubmissionsColumn = !userIsEndUser;
```
rocio-desantiago, line 22: is `!userIsEndUser` is it `userIsAdmin`?

rocio-desantiago (commented):
This has already been merged but I still have a lingering question.

PR #218, opened 2025-04-22
Rename Radio.value to Radio.choices

benmartin-coforma:
### Description
This PR updates the shape of RadioTemplate and ReportingRadioTemplate, renaming the array of ChoiceTemplates from `value` to `choices`.

Why?
For one, this is a more appropriate name. To me, "value" sounds like the (string) value of an individual option, or perhaps the (string) value of the option of the radio button group that is currently selected. This property is neither; it is an array of all the options in the group, and in fact we already had some code iterating over `for (const choice of element.value)`. That code now reads "for const choice of element dot choices" ✨ 

For two, I suspect that the value property of radio button groups was somehow being overwritten, sometimes. Leading to intermittent bugs. None of the places I've renamed it are assigning to it, so _if_ I am correct and that was the cause of our bugs, _then_ with this rename I have "fixed" the bugs. This would leave open some questions, such as:
   * Was the code writing to radio.value doing anything useful? Won't we be missing that functionality now?
   * If the code writing to radio.value was not doing anything useful, shouldn't we find & delete that code? Where is it, anyway?
   * What would have made those bugs intermittent?
   * If I am wrong, then what _is_ the cause of our intermittent bugs?

These are excellent questions. I propose we address them later; I believe this rename is worthwhile for its own sake, and if it makes our app work more smoothly then so much the better. We can leave dead code cleanup for later.

### Related ticket(s)
CMDCT-4542

---
### How to test
https://dfmg71ede69ah.cloudfront.net/

This affects every radio button in the app, including the special "are you reporting on this measure" buttons. So,
1. Log into the app as a state user
2. Create a new report
   * Note that the report is created successfully
3. Click around and change some radio button values
   * Note that the radio options render correctly
   * Note that child inputs render correctly, if certain radio buttons are checked (for example, the "by who" textbox if you say the measure was audited)
4. Refresh the page
   * Note that the values you checked remain checked after reloading
5. ~Verify that we don't have a complicated validation regression:~
   ~1. Completely fill out a measure page, including a radio button with a child input.~
   ~2. Note that the "complete measure" button is enabled.~
   ~3. Delete the value of the child input.~
   ~4. Note that the input is now red, saying an answer is required.~
   ~5. Note that the "complete measure" button is now disabled.~
   ~6. Select the _other_ radio option (for example, say that the measure was not audited)~
   ~7. Note that the child input with the error disappears~
   ~8. _And, crucially_, note that the "complete measure" button is re-enabled. If it is still disabled, then this PR broke a previous fix for this scenario.~
Edit: Step 5 won't work, for unrelated, pre-existing reasons.

### Important updates
Note that, since this PR changes the report templates, reports created prior to the merge of this PR will need to be deleted & re-created.

---
### Author checklist

- [ ] I have manually tested this PR in the deployed environment


BearHanded (approved):
This appears to work, and is sane, and makes it more readable. It might have merge conflicts with my current branch depending on who gets in first? Uncertain.

rocio-desantiago (approved):
Tested in deploy ✅ 
Wow so many choices 🥁

rocio-desantiago (approved):
More choices!

ajaitasaini (approved):

PR #231, opened 2025-04-24
QMS Content Update - Measure Overview

ajaitasaini:
### Related ticket(s)

CMDCT-

---
### Important updates

---
### Author checklist

---

### DIFF
--- a/services/app-api/forms/2026/elements.ts
+++ b/services/app-api/forms/2026/elements.ts
@@ -51,7 +51,10 @@ export const measureInstructions: AccordionTemplate = {
   id: "measure-instructions",
   label: "Instructions",
   value:
-    "[Optional instructional content that could support the user in completing this page]",
+    "<b>Instructions for Completing this Measure</b>" +
+    "<br><br>Before you can click the <b>“Complete measure”</b> button, you must answer all required (non-optional) questions for the measure and any associated measure sections (such as delivery method or measure part)." +
+    "<br><br>Please review your responses to ensure all mandatory fields are filled out before proceeding." +
+    "<br><br>The <b>“Clear measure data”</b> button can be used to reset the entire measure (including any completed sections).  All data previously entered will be cleared and not submitted upon report completion.",
 };
 
 export const measureDetailsSection: MeasureDetailsTemplate = {
@@ -84,9 +87,9 @@ export const isTheStateReportingThisMeasure: RadioTemplate = {
     "Warning: Changing this response will clear any data previously entered in this measure.",
   id: "measure-reporting-radio",
   value: [
-    { label: "Yes, the state is reporting on this measure", value: "yes" },
+    { label: "Yes, the state is reporting on this measure.", value: "yes" },
     {
-      label: "No, CMS is reporting this measure on the state's behalf",
+      label: "No, CMS is reporting this measure on the state's behalf.",
       value: "no",
     },
   ],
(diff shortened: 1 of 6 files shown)


services/app-api/forms/2026/elements.ts
```diff
@@ -51,7 +51,10 @@ export const measureInstructions: AccordionTemplate = {
   id: "measure-instructions",
   label: "Instructions",
   value:
-    "[Optional instructional content that could support the user in completing this page]",
+    "<b>Instructions for Completing this Measure</b>" +
+    "<br><br>Before you can click the <b>“Complete measure”</b> button, you must answer all required (non-optional) questions for the measure and any associated measure sections (such as delivery method or measure part)." +
+    "<br><br>Please review your responses to ensure all mandatory fields are filled out before proceeding." +
+    "<br><br>The <b>“Clear measure data”</b> button can be used to reset the entire measure (including any completed sections).  All data previously entered will be cleared and not submitted upon report completion.",
```
benmartin-coforma, line 57: Nitpicks here:
* I'd rather see us mark blocks of text with `<p>` tags, than build them out with manual linebreaks.
* Use of the `<b>` tag to boldify text is generally discouraged in favor of `<strong>` when the reason for bolding is to indicate importance, so I recommend it for the subheading here.
   * Although arguably this should be a `<h2>` or `<h3>` instead, since it is literally a subheading? But I won't get fussy about that.
* On the other hand `<b>` _can_ be used to bring attention to keywords, and I think calling out specific buttons' text qualifies. So I'm not recommending to swap out all uses here.
* And this can turn into a hair-splitting argument about the line between styling and semantics. You can [read all about it here.](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/b#usage_notes)

```suggestion
    "<strong>Instructions for Completing this Measure</strong>" +
    "<p>Before you can click the <b>“Complete measure”</b> button, you must answer all required (non-optional) questions for the measure and any associated measure sections (such as delivery method or measure part).<p>" +
    "<p>Please review your responses to ensure all mandatory fields are filled out before proceeding.</p>" +
    "<p>The <b>“Clear measure data”</b> button can be used to reset the entire measure (including any completed sections).  All data previously entered will be cleared and not submitted upon report completion.</p>",
```

services/app-api/forms/2026/elements.ts
```diff
@@ -156,6 +165,8 @@ export const didYouFollowSpecifications: RadioTemplate = {
           id: "measure-following-tech-specs-no-explain",
           label: "Please explain the variance.",
           required: true,
+          helperText:
+            "Include the name of which technical specifications that were used in the reporting of this measure, or any data elements that were collected outside of the most current guidance (e.g. sampling size, population, denomination calculation etc.)",
```
benmartin-coforma, line 169: This does match the wording in the document, but I think the wording in the document is probably incorrect, because it's ungrammatical. Here's one way we might fix it:

```suggestion
          helperText:
            "Include the name of which technical specifications were used in the reporting of this measure, or any data elements that were collected outside of the most current guidance (e.g. sampling size, population, denomination calculation etc.)",
```

benmartin-coforma (dismissed):
Code looks good! Thanks Ajaita 👍

rocio-desantiago (dismissed):
Looks good

rocio-desantiago (approved):
Good catch 👍

benmartin-coforma (approved):

PR #233, opened 2025-04-24
Sidebar Show/Hide Fix + Unlock Modal + Spinners

ailZhou:
### Description

1) Sidebar show & hide is now part of the state management. It will stay hidden when you go into a full page view and then go back to the sidebar view.

2) Missing unlock modal 

- There is this modal now that pops up when you unlock a report.

Screenshot 2025-04-29 at 3 51 43 PM

3) Spinners (all these are subjected to change since Megan is out)

- There are now spinners when you click submit report 

https://github.com/user-attachments/assets/9c027c9b-7e88-4019-ae53-bec51be0d8f7

- A spinner when you Archive and a spinner when you click Unlock

https://github.com/user-attachments/assets/3a4ec9f2-b925-4ab9-a91d-e299158d1c62

### Related ticket(s)

CMDCT-4566

---
### How to test

Deploy Link: https://d1cy49o5i9piuq.cloudfront.net/

Review that the changes above exist and that there's no unexpected weirdness.

### Important updates

---
### Author checklist

---


BearHanded (dismissed):

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -39,25 +56,14 @@ export const getStatus = (report: Report) => {
...
-  showEditNameColumn: boolean | undefined,
-  showReportSubmissionsColumn: boolean,
-  showAdminControlsColumn: boolean | undefined,
-  openAddEditReportModal: Function,
-  navigate: NavigateFunction,
-  userIsEndUser: boolean | undefined,
-  toggleArchived: Function,
-  toggleRelease: Function
-) => {
+export const HorizontalTable = (props: TableProps) => {
```
benmartin-coforma, line 59: Nice! I'm glad that, when faced with adding an 11th and 12 argument to this function, you instead collapsed them all down into a single props object. Would you consider re-expanding the function signature so that the props are immediately destructured into named variables? It's more lines of code, but personally I find it easier to trace individual properties' usage. Or is having fewer lines of code the exact reason you made the change?

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -27,6 +28,22 @@ import { useState } from "react";
...
+  showEditNameColumn: boolean | undefined;
+  showReportSubmissionsColumn: boolean;
+  showAdminControlsColumn: boolean | undefined;
+  openAddEditReportModal: Function;
+  navigate: NavigateFunction;
+  userIsEndUser: boolean | undefined;
+  toggleArchived: Function;
+  toggleRelease: Function;
+  archiving: number | undefined;
+  unlocking: number | undefined;
```
benmartin-coforma, line 46: This is very neat. It took me a minute to understand why these are numbers, but now I see: they're indexes into the table rows. With these indexes, we can display the spinner in exactly the right spot.

Do you think there's a way we could make that more obvious? Maybe with a `/** doc comment*/`, or with a rename? Maybe something like `archivingIndex`?

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -75,38 +81,49 @@ export const HorizontalTable = (
...
                   variant="link"
-                  onClick={async () => await toggleRelease(idx)}
+                  fontSize="sm"
+                  onClick={async () => await props.toggleRelease(idx)}
                   disabled={
-                    report.status !== ReportStatus.SUBMITTED || report.archived
+                    report.status !== ReportStatus.SUBMITTED ||
+                    report.archived ||
+                    props.unlocking === idx
                   }
```
benmartin-coforma, line 109: I'm not sure about this - it may be a good idea for us to instead disable the button when `props.unlocking !== undefined`

My thinking is that, if the user clicks two unlock buttons quickly enough, the `props.unlocking` value of the first will be overwritten by the second. I don't _think_ this is a problem as far as correctness is concerned; they're independent requests that will parallelize just fine. But we will get confused about where to display the spinner(s). So we should prevent the user from clicking any unlock button until the first unlock action is complete.

Does that sound right, or am I off base?

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -118,26 +135,16 @@ export const HorizontalTable = (
...
-  showEditNameColumn: boolean | undefined,
-  showReportSubmissionsColumn: boolean,
-  showAdminControlsColumn: boolean | undefined,
-  openAddEditReportModal: Function,
-  navigate: NavigateFunction,
-  userIsEndUser: boolean | undefined,
-  toggleArchived: Function,
-  toggleRelease: Function
-) => {
+export const VerticleTable = (props: TableProps) => {
```
benmartin-coforma, line 138: While you're in here I would absolutely love if you renamed this component to `VerticalTable` 🤠

services/ui-src/src/components/report/SubmitReportModal.tsx
```diff
@@ -4,6 +4,7 @@ export const SubmitReportModal = (onClose: Function, onSubmit: Function) => {
   const submitHandler = async () => {
     await onSubmit();
   };
+
```
benmartin-coforma, line 7: Yesss give the code room to breathe 🙏

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
ailZhou: i did consider it but then it felt like why did i bother merging them into a single prop object if i was doing to destructured them into variables right after 😓

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
ailZhou: i did think about that as a concern, but i don't think from a UX perspective that it makes sense. Also, these spinners are more of a placeholder until we have time to think the scenario though so it's not a one and done situation. We have questions like, do we allow users to do multiple archiving/unlocks at once? Is that something we should handle? etc.

rocio-desantiago (approved):
All actions work accordingly.

Side note: not related to the changes in this PR but when I selected "No" for the poms question, the poms still showed up. I was trying to fill less out to submit but that plan didn't work. Might need a spike to investigate that.

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
benmartin-coforma: Alright. I'm fine leaving "fix up spinner behavior when archiving multiple reports as once" as a potential future ticket.

benmartin-coforma (approved):
Thanks Aileen!

PR #239, opened 2025-04-28
[CMDCT-4567] Report Data Deduplication

BearHanded:
### Description

This should be pretty reviewable by commit
1) Shift the types around a bit to remove the measureLookup and templates. This info gets folded into the pages as they are built, and is practically a x2 multiplier on data size.
2) Remove "title" field, as "name" was being conventionally used. Title was only interacted with in tests and on the Export page. Unified.

No fancy screenshots to post, everything should behave as normal.

### Related ticket(s)

CMDCT-4567

---
### How to test

https://d7tpl59j331fb.cloudfront.net/
- Build a report
- Inspect the network response, make sure measureLookup and measureTemplates aren't in there, just .pages.
- Check that the more complex pages of the app are looking that use that information built into pages. The delivery methods table is a good example.

### Important updates

---
### Author checklist

---


services/app-api/handlers/reports/buildReport.ts
benmartin-coforma, line 1: I promised you an opinion on this: I do think we end up at a cleaner place if we build the report up from the pieces of the templates, versus cloning the template and deleting the pieces we don't need. Would look something like this:

```ts
//...
) => {
  const year = reportOptions.year;
  const template = getReportTemplate(reportType, year);
  const cmitList = getCmitInfo(year);

  const report: Report = {
    type: template.type,
    year: template.year,
    state: state,
    id: KSUID.randomSync().string,
    created: Date.now(),
    lastEdited: Date.now(),
    lastEditedBy: user.fullName,
    lastEditedByEmail: user.email,
    status: ReportStatus.NOT_STARTED,
    name: reportOptions.name,
    options: reportOptions.options,
    archived: false,
    submissionCount: 0,
    pages: structuredClone(template.pages),
  };

  if (reportType == ReportType.QMS) {
    // ...
```

benmartin-coforma (approved):

rocio-desantiago (approved):
Confirmed changes in deploy link

PR #212, opened 2025-04-18
[HCBS] State Performance Target Question Required

rocio-desantiago:
### Description

This PR fixes the bug where complete section was showing up even though the answer for the "required" performance Target question was left blank.

Alternate option for the logic is: 
Looks cleaner but not sure if as readable. Curious what thoughts are:
```
if (
  [uniqueRate.rate, uniqueRate.performanceTarget].some(
    (uniqueRateValue) => uniqueRateValue === undefined || uniqueRateValue === ""
  )
) {
  return false;
}
```

### Related ticket(s)

CMDCT-4554

---
### How to test

**Deploy link:** https://d13wce1kd9m1xw.cloudfront.net/

1. Sign into HCBS, any state user.
2. Create a new QMS report, say yes to the POM measures to check those too.
3. Go to any measure and confirm that if you leave the answer blank for the question: "What is the 2028 state performance target for this assessment?", that the complete button stays disabled. To test this accurately, make sure to click anywhere on the page due to onblur behavior.
4. 

---
### Author checklist

- [x] I have performed a self-review of my code
- [ ] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
- [ ] I have updated relevant documentation, if necessary
---

_convert to a different template: [test → val](?expand=1&template=test-to-val-deployment.md)_ | _[val → prod](?expand=1&template=val-to-prod-deployment.md)_

PR #244, opened 2025-04-30
NDR Field Bug Fix

ailZhou:
### Description

[Same issue as this PR](https://github.com/Enterprise-CMCS/macpro-mdct-hcbs/pull/212) but with NDR Field rate calcs. The solution is to also see if performanceTarget or rates is "".

https://github.com/user-attachments/assets/014e5493-fbc3-413a-b908-7339546653ad

Also in this pr is a styling fix, I moved the button back to the right:
Screenshot 2025-04-30 at 2 42 56 PM

**Why does the solution require a cast?**
`(uniqueRate.rate as number | string) === ""`

So typescript is working correctly, as we have set rate to be a number and a number only. The issue is that for user experience, we have to be able to have an empty textbox and an undefined was not enough. An undefined causes the input to not fire our onchange/onblur correctly but we didn't want to allow it to take strings as we only wanted numbers to be save so we have this weird amalgamation. This is why we have to case before checking.

### Related ticket(s)

---
### How to test

2) Create a new QMS report, no POM measures needed
3) Go to Optional Measure Results page
4) Edit "HCBS-10: Self-direction of Services and Supports Among Medicaid Beneficiaries Receiving LTSS through Managed Care Organizations" 
5) Fill out a performance measure and make sure that the [Complete section] button is enabling/disabling correctly.

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/utils/state/reportLogic/completeness.ts
+++ b/services/ui-src/src/utils/state/reportLogic/completeness.ts
@@ -200,8 +200,16 @@ const rateIsComplete = (element: PerformanceRateTemplate) => {
     for (const rateAnswer of element.answer) {
       if (!rateAnswer.rates) return false;
       for (const uniqueRate of rateAnswer.rates) {
-        if (uniqueRate.performanceTarget === undefined) return false;
-        if (uniqueRate.rate === undefined) return false;
+        if (
+          uniqueRate.performanceTarget === undefined ||
+          (uniqueRate.performanceTarget as number | string) === ""
+        )
+          return false;
+        if (
+          uniqueRate.rate === undefined ||
+          (uniqueRate.rate as number | string) === ""
+        )
+          return false;
       }
     }
   }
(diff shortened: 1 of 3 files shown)


services/ui-src/src/utils/state/reportLogic/completeness.ts
```diff
@@ -200,8 +200,13 @@ const rateIsComplete = (element: PerformanceRateTemplate) => {
...
       if (!rateAnswer.rates) return false;
       for (const uniqueRate of rateAnswer.rates) {
-        if (uniqueRate.performanceTarget === undefined) return false;
-        if (uniqueRate.rate === undefined) return false;
+        if (
+          uniqueRate.performanceTarget === undefined ||
+          (uniqueRate.performanceTarget as number | string) === "" ||
+          uniqueRate.rate === undefined ||
+          (uniqueRate.rate as number | string) === ""
+        )
```
BearHanded, line 208: Alternatively something like this might be more readable
```
const emptyAnswers = ["", undefined];
if (
  emptyAnswers.includes(uniqueRate.performanceTarget) || 
  emptyAnswers.includes(uniqueRate.rate))
) {
```

BearHanded (approved):

services/ui-src/src/utils/state/reportLogic/completeness.ts
ailZhou: change made, using that style instead

BearHanded (approved):

benmartin-coforma (approved):
Seems fine to me

PR #245, opened 2025-04-30
[CMDCT-4617] Radio Tab Fix

BearHanded:
Tabbing through radios was having a weird effect when you skipped right past. This was due to removing the onBlur handling in favor of onchange, but KEEPING the onBlur handlers in the component.

The component would write the radio value of the previewed item into the hook form on tab, whatever we had in place at the wrapper level before prevented this for empty values.

This removes the vestigial traces of the blur handling, and assumes we're using change for everything.

I tested most odd interactions I could think of, but please poke around, give it a shot. 
### Related ticket(s)

CMDCT-4617

---
### How to test

Report Creation Modal, 
- Enter name
- Tab through to submit without filling anything out
- You shouldn't be able to submit, good job us.

Measure Details
- Tab through the delivery method question to complete
- the table should not have lit up an edit link unexpectedly

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/fields/DateField.test.tsx
+++ b/services/ui-src/src/components/fields/DateField.test.tsx
@@ -64,18 +64,12 @@ describe("<DateField />", () => {
       render(dateFieldComponent);
       const dateFieldInput = screen.getByRole("textbox");
 
-      await userEvent.type(dateFieldInput, "10162024[Tab]");
-
-      /*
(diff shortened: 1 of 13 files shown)


benmartin-coforma (approved):
I haven't tested this, but the code looks A-OK

rocio-desantiago (approved):
Tested and it's being normal 😄

PR #253, opened 2025-05-09
Let things be a little dirtier

benmartin-coforma:
### Description
1. We sanitize the body of every request that hits our API, to ensure we never store or even interact with potentially dangerous data.
2. We use the library DOMPurify for this, with a very strict config set to allow through only a limited set of HTML tags.
3. Within our own report templates, we use tags outside of that limited set.
4. As a user completes a report, we store the answers directly in the template. In other words, every time a user saves report data, they are also re-saving the template.
5. Therefore, the first time each report is saved by a user, that report's template is mangled by DOMPurify.

This PR loosens up the DOMPurify config, to allow a slightly broader set of tags, which includes all tags currently in use in our report templates.

In order to verify this is working correctly, and to ensure future edits to the report templates do not quietly add unsanitizable tags, I created a unit test that actually instantiates all possible versions of the QMS report, and asserts that each version makes it through the sanitizer unchanged.

In order to write that unit test, I had to separate the action of building a report from the action of saving that report to the database. Well, I guess I didn't _have_ to, but it saved me from some extra mocking, and it makes more sense to me this way.

Note that I had to make some edits to the instructions in order to make sure DOMPurify wouldn't change them. It was doing weird stuff like generating closing `<p>` tags in weird spots, and stripping out empty ones. It's possible these edits changed the spacing on the page. Hopefully it still looks right?

image
image
image
image

### Related ticket(s)
CMDCT-4630

---
### How to test
https://d1cx8o2bjaw6wo.cloudfront.net/

I guess, make sure you can still create a report.

And save changes to it.

And that the instruction accordions look OK, I guess.

That should be it though.

### Important updates
n/a

---
### Author checklist

### DIFF
--- a/services/app-api/forms/2026/elements.ts
+++ b/services/app-api/forms/2026/elements.ts
@@ -58,19 +58,20 @@ export const measureInstructions: AccordionTemplate = {
   label: "Instructions",
   value:
     "<strong>Instructions for Completing this Measure</strong>" +
-    "<p>Before you can click the <b>“Complete measure”</b> button, you must answer all required (non-optional) questions for the measure and any associated measure sections (such as delivery method or measure part).<p>" +
+    "<p>Before you can click the <b>“Complete measure”</b> button, you must answer all required (non-optional) questions for the measure and any associated measure sections (such as delivery method or measure part).</p>" +
     "<p>Please review your responses to ensure all mandatory fields are filled out before proceeding.</p>" +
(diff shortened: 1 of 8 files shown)


rocio-desantiago (approved):
Report still working ✅

ailZhou (approved):
lgtm! no issues with creating reports and text looks the same

PR #254, opened 2025-05-12
[Cmdct-4610] Create and Access TA & CI reports

BearHanded:
### Description

Adds functionality for the TA & CI reports, plugging into the existing tables.

- Adds templates for each report type, moving each report into a year and type folder
- Ties those new forms into the formSelection method in the api, at that point most API functionality works
- Changes the reportType QMS check in build to instead look for if it is a report with a measureConfig. The new reports are simpler and don't do that, and we may reuse that "measureConfig" logic for other reports.
- Restructures the AddEditReportModal to pull it's verbiage and options from other files per report type, letting us reuse that component and make changes in one place per report type.
- Components like SubNavHeader and Sidebar that display report type verbiage now key off type to get their displayed value. If this trend expands, we can make actual verbiage files, but the AddEditReportModal variant is slightly more complicated. Not a lot of reuse just yet.

TL;DR You can create the new report types, access them, edit them, probably complete them. 

image
image

### Related ticket(s)

CMDCT-4610, CMDCT-4611, CMDCT-4612

---
### How to test

https://d3urd1ouex0cuw.cloudfront.net/
Create a CI Report
- The options should differ from QMS creation
- Report can be renamed after creation
- Clicking Edit takes you into the report
- The report can be edited and saved

Create a TA Report
- repeat the above

### Important updates

---
### Author checklist

---


services/ui-src/src/components/layout/Header.tsx
benmartin-coforma, line 1: Suggestion: This is unrelated to your current PR, but would it be a huge lift to have `SubnavBar` call `const { state, reportType } = useParams()`? As opposed to slicing up the path ourselves & passing those values in through props.

services/ui-src/src/components/modals/AddEditReportModal.test.tsx
```diff
@@ -177,6 +178,37 @@ describe("Test submit", () => {
...
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+  test.each([
+    { type: ReportType.QMS, text: "Quality Measure Set Report" },
+    { type: ReportType.TA, text: "TACM Report" },
+    { type: ReportType.CI, text: "CICM Report" },
+  ])("$type report type renders a title", ({ type, text }) => {
+    screen.debug();
```
benmartin-coforma, line 194: Nit: we won't need this ~console.log~ I mean, `screen.debug()`

```suggestion
```

services/ui-src/src/components/modals/AddEditReportModal.tsx
```diff
@@ -1,10 +1,32 @@
...
+  selectedReport: Report | undefined
+) => {
+  switch (reportType) {
+    case ReportType.QMS:
+      return QmsOptions(selectedReport);
+    case ReportType.TA:
+      return TaOptions();
+    case ReportType.CI:
+      return CiOptions();
+  }
```
benmartin-coforma, line 28: Question: Would an `assertExhaustive(reportType)` make a difference here?

I mean, if you delete a case, does the switch already complain that it's not exhaustive? If not, then I think an assertion is called for. If so, that's wonderful, and I have a bunch of places I want to go delete such assertions.

services/ui-src/src/components/report/Sidebar.tsx
```diff
@@ -14,6 +15,18 @@ const navItem = (title: string, index: number) => {
...
+const getTitle = (report: Report) => {
+  if (!isReportType(report.type)) return "";
+  switch (report.type) {
+    case ReportType.CI:
+      return "CICM Report";
+    case ReportType.QMS:
+      return "Quality Measure Set Report";
+    case ReportType.TA:
+      return "Timely Access Report";
+  }
```
benmartin-coforma, line 27: Same question here regarding assertExhaustive

benmartin-coforma (commented):
I don't see any major problems here.

ajaitasaini (dismissed):
tested locally, works as expected. functionality is correct + correct URLs are being reached. Don't see any glaring issues with the code!

services/ui-src/src/components/modals/AddEditReportModal.test.tsx
BearHanded: oops yes

services/ui-src/src/components/layout/Header.tsx
BearHanded: I think it just recalls moving the route provider or whatever it's called up a level or two.

ajaitasaini (approved):

services/ui-src/src/components/layout/Header.tsx
benmartin-coforma: Okey doke. That's a potential drive-by refactor sometime, then.

benmartin-coforma (approved):
Code looks good to me!

PR #255, opened 2025-05-12
CMDCT-4629 post cdk conversion cleanup and removal of all serverless references

peoplespete:
### Description
Now that HCBS is on CDK we can rip out all the old references to serverless and throw a couple other enhancements we've found out as we've done #3 and #4 cdk conversions.

### Related ticket(s)
CMDCT-4629

---
### How to test
App should still work:
https://d13gl52oxjf92t.cloudfront.net/

### Important updates
NA

---
### Author checklist

---


BearHanded (dismissed):

deployment/stacks/api.ts
```diff
@@ -116,9 +107,6 @@ export function createApiComponents(props: CreateApiComponentsProps) {
       tables.map((table) => [`${table.id}Table`, table.name])
     ),
     BOOTSTRAP_BROKER_STRING_TLS: brokerString,
-    COGNITO_USER_POOL_ID: userPoolId || process.env.COGNITO_USER_POOL_ID,
```
JonHolman, line 119: I thought we added this to get the parameters out of SSM, we don't need this anymore?

peoplespete: Because the api already does the authentication so checking that the api token matches the cognito params is redoing on every single request what cognito just did before it let the request into the lambda. QMR was already set up like this.

deployment/stacks/parent.ts
```diff
@@ -61,49 +45,27 @@ export class ParentStack extends Stack {
...
 
-    if (isLocalStack) {
-      createApiComponents({
-        ...commonProps,
-        tables,
-        vpc,
-        kafkaAuthorizedSubnets,
-      });
-      /*
-       * For local dev, the LocalStack container will host the database and API.
```
JonHolman, line 72: Ben wanted to add this note for SEDS.  I think we should keep it.  Thoughts?

JonHolman (dismissed):

deployment/stacks/parent.ts
peoplespete: Fixed.

PR #262, opened 2025-05-14
Better labeling for editing and entering reports

ajaitasaini:
### Description

Modifying aria-labels for the "Edit Report Name" and "Edit" buttons in the report dashboard to be more specific
image

### Related ticket(s)

CMDCT-4466

---
### How to test

**Deploy Link:** https://d219no0jteicvi.cloudfront.net/

1. Create a report, and inspect the aria-label "Edit Report Name" and "Edit" buttons. 
2. Confirm that they are now report name specific (e.g. if the report is named abc the aria labels should be "Edit abc report name" and "Edit abc report"

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
@@ -66,7 +66,11 @@ export const HorizontalTable = (props: TableProps) => {
           {props.showEditNameColumn && (
             <Td fontWeight={"bold"}>
               <button onClick={() => props.openAddEditReportModal(report)}>
-                <Image src={editIcon} alt="Edit Report Name" minW={"1.75rem"} />
+                <Image
+                  src={editIcon}
+                  aria-label={"Edit " + report.name + " report name"}
+                  minW={"1.75rem"}
+                />
               </button>
             </Td>
           )}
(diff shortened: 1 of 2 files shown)


rocio-desantiago (approved):
Inspected and saw the changes in the console for the two edit buttons

cy-by (approved):
The buttons have accessible names that I expected to see!

hcbs-updated-accessible-name

Nitpick: The `<image>` `alt` attribute didn't need to be changed to `aria-label` as in this case, it does the same thing. But it doesn't hurt anything, so no need to for further action, just a FYI.

rocio-desantiago:
> The buttons have accessible names that I expected to see!
> 
> hcbs-updated-accessible-name
> 
> Nitpick: The `<image>` `alt` attribute didn't need to be changed to `aria-label` as in this case, it does the same thing. But it doesn't hurt anything, so no need to for further action, just a FYI.

I am obsessed with this image. What tool did you use to do that type of focusing?

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -66,7 +66,11 @@ export const HorizontalTable = (props: TableProps) => {
           {props.showEditNameColumn && (
             <Td fontWeight={"bold"}>
               <button onClick={() => props.openAddEditReportModal(report)}>
-                <Image src={editIcon} alt="Edit Report Name" minW={"1.75rem"} />
+                <Image
+                  src={editIcon}
+                  aria-label={"Edit " + report.name + " report name"}
```
benmartin-coforma, line 71: I've got an annoying nit here - would you mind swapping out the string concatenation for string interpolation? I try to stick to interpolation because there are cases where it's a lot more readable.

```suggestion
                  aria-label={`Edit ${report.name} report name`}
```

ajaitasaini: good suggestion, thank you!

cy-by (approved):

benmartin-coforma (approved):

PR #264, opened 2025-05-14
Remove need to cast the props of every element component

benmartin-coforma:
### Description
This PR refactors `Page.tsx` and `PageElementProps` such that every page element component can require a specific subtype of `PageElement`. The HeaderElement knows it will get a HeaderTemplate, the ParagraphElement knows it will get a ParagraphTemplate, the [SnozzElement knows it will get a SnozzTemplate](https://www.youtube.com/watch?v=xDbJhqyuV1M)

This change has three parts:
1. The `PageElementProps` interface has been made generic. When referencing it, you can tell it what type you want `props.element` to be.
2. All of the individual components pass their specific template types to that generic interface. Thus Typescript will now enforce those specific types.
3. Page.tsx has been reworked so that Typescript can do the necessary type inference. It's still functionally doing the same work as before, it's just structured to be easier for easier Typescript to analyze.
   - I don't love having six dots in a row for `<Foo {...{...props, element}}/>`
   - But I do like having the messiness contained in here, rather than scattering `as FooTemplate` casts around every component.

### Related ticket(s)
n/a

---
### How to test
https://d1uh8l767vxupq.cloudfront.net/

The app should work as before; no behavioral changes.

### Important updates
n/a

---
### Author checklist


benmartin-coforma:
The list of commits is longer than it feels like it should be, because I branched off `more-typing` before it was merged into main, so I have all of its individual commits and not the squash commit from when that PR was merged. The diffs are correct though!

ajaitasaini (approved):

PR #265, opened 2025-05-14
Remove some more Typescript-confounding code

benmartin-coforma:
### Description
We were using `[key: string]: any;` as a catch-all pass-through prop for some of our components. This PR removes those.

We still had the parseCustomHtml code from MCR / MFP, even though HCBS report templates use actual HTML. This PR removes it.

### Related ticket(s)
n/a

---
### How to test
https://d3bb4qkks39r4j.cloudfront.net/

There should be no behavioral changes.

### Important updates
n/a

---
### Author checklist


rocio-desantiago (approved):
Nice cleanup on the `data-testid`'s and from `sx` to `sxOverride` too!

ajaitasaini (approved):

services/ui-src/src/components/cards/Card.tsx
benmartin-coforma, line 1: * None of the places we use `<Card>` in the app were passing anything but `children`, so I could remove the `[key: string]: any` with no repercussions
* The only thing we're doing with `children` is passing it through to `<Box>`, which will handle it just fine implicitly
* The only thing we're setting on the Box is `sx`, so we can just define our props interface as "Anything you would give to a Box, except `sx`" (which is what `Omit` does here)
* If you want to use a `<Card>` but also want to give it an `sx` prop, then you can just use a `<Box>` instead.

PR #267, opened 2025-05-19
[HCBS] - Make sure start date is before end date in banner form

rocio-desantiago:
### Description

There is a lot of yup validation here -- as Angela suggested. 
After Ben's awesome re-factor, typing the date correctly was possible!
Note: Dates are so weirrddddd. We go from `string` to `number` to "Date" to "ET" to "Utc" to `epoch` to `eons` to `galaxy years` to `1970` because 1970 is when time started - I'm just kidding - I'm exaggerating here.

Here's some context for some of the yup updates:
* yup quirk: `startDate` is passed into the custom validation function as a singleton array, instead of just its actual value. Thanks Ben for adding the code that => `always expects startDate to be an array with one number in it and the test for that on the frontend`
* `schema.test({...})` defines a custom validation rule for the endDate field, ensuring that if both startDate and endDate are present AND numbers, endDate must be after startDate. If not, it triggers an error message (error.END_DATE_BEFORE_START_DATE).
* `.when("startDate", (startDate, schema) => { ... })` - The first argument, "startDate", specifies which field to watch for changes. The second argument is a function that receives the value of startDate and the current schema for endDate. If startDate is a number, the schema adds a custom test to ensure that endDate is after startDate. If startDate is not a number (or not present), it just keeps the schema as-is (no extra rule).
* `checkDateCompleteness` is no longer needed as `parseMMDDYYYY` covers what it was doing. Twas redundant. Date parsing has evolved. For context - checkDateCompleteness was simply 
* for consistency in some unit tests instead of `test(` -> `it(`
* maybe too many time.ts tests, but was trying to bring code coverage up

### Related ticket(s)

CMDCT-4301

---
### How to test
**Deploy Link:** https://d2mzrklpqm7rn2.cloudfront.net/
1. Login as an admin user: adminuser@test.com
2. Go to top right and click on "My Account" -> "Manage Account"
3. Click on "Banner Editor"
4. Enter a date for "Start date"
5. Enter an older for "End date"
6. The validation error matches MCR's version: "End date can't be before start date"
    image of validation error when End date is older than Start date

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/utils/other/time.test.ts
+++ b/services/ui-src/src/utils/other/time.test.ts
@@ -9,6 +9,7 @@ import {
   convertDateUtcToEt,
   formatMonthDayYear,
   getLocalHourMinuteTime,
+  parseMMDDYYYY,
   twoDigitCalendarDate,
 } from "./time";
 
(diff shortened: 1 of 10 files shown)


services/ui-src/src/utils/other/time.test.ts
```diff
@@ -143,27 +124,105 @@ describe("utils/time", () => {
...
+    const date = parseMMDDYYYY("12/25/2023");
+    expect(date).toBeInstanceOf(Date);
+    expect(date?.getFullYear()).toBe(2023);
+    expect(date?.getMonth()).toBe(11); // December (0-indexed)
+    expect(date?.getDate()).toBe(25);
+    expect(date?.getHours()).toBe(0); // Should be normalized to midnight
+    expect(date?.getMinutes()).toBe(0);
+    expect(date?.getSeconds()).toBe(0);
+    expect(date?.getMilliseconds()).toBe(0);
+  });
```
benmartin-coforma, line 163: This is a very nice test. Just a thought here, and not a suggestion:

Typescript doesn't understand that `expect(date).toBeInstanceOf(Date)` will throw an exception if `date` is null (or undefined). Which is why it's asking for all these `?.` operators.

Typescript _would_ understand `assert(date instanceOf Date)`, if we were to `import assert from "node:assert"`. So if we swapped that out, we could drop all of these question marks. Which feels a little better, because it truly does not make sense for us to have them here. As you and I read the code, we can clearly see that none of them will trigger, because `date` will always be defined.

On the other hand, if `parseMMDDYYYY` _were_ to break, and return null (or undefined) for this input, we would probably get a nicer error message out of `expect` than `assert` - because "nice error messages" is precisely the thing that `expect` was designed for. So swapping it out, we would be losing a little something.

On the third hand, how much nicer would that error message be really? (This is a question that you or I could answer in a minute by making the switch, breaking the function, and re-running the test). And, how likely is the function to ever break in that way? So would we be keeping the silly `?.` operators around for a marginal benefit in a scenario that will probably never happen?

On the fourth hand, we are paying a marginal cost by thinking & talking about this right now, so maybe a change would not be worth it.

I'm out of hands & feet so that's gotta be the end of the thought. It's up to you to write the code how you like :)

services/ui-src/src/utils/other/time.test.ts
```diff
@@ -143,27 +124,105 @@ describe("utils/time", () => {
...
+    const date = parseMMDDYYYY("02/29/2028"); // 2028 is a leap year
+    expect(date).toBeInstanceOf(Date);
+    expect(date?.getFullYear()).toBe(2028);
+    expect(date?.getMonth()).toBe(1); // February (0-indexed)
+    expect(date?.getDate()).toBe(29);
+  });
+
+  it("should return null for 02/29 in a non-leap year", () => {
+    const date = parseMMDDYYYY("02/29/2027");
+    expect(date).toBeNull();
```
benmartin-coforma, line 183: I just want to point out that `.toBeUndefined()` exists, if you end up changing that. And `.toBeFalsey()` exists too, if you want these tests to pass regardless of that implementation detail. Not a suggestion, just a thought.

services/ui-src/src/utils/other/time.test.ts
benmartin-coforma, line 1: These are really outstanding tests, and I appreciate you going in to change the other test assertions' phrasing. They read much more smoothly now. 😍

services/ui-src/src/utils/validation/bannerValidation.ts
```diff
@@ -1,59 +1,106 @@
...
   description: string().required(),
   link: string().url().notRequired(),
   startDate: number().notRequired(),
-  endDate: number().notRequired(),
+  endDate: number()
+    .notRequired()
+    .when("startDate", (startDate, schema) => {
+      const actualStartDate = Array.isArray(startDate)
+        ? startDate[0]
+        : startDate;
```
benmartin-coforma, line 89: I don't understand this part. If I'm following correctly, `startDate` is set on lines 31-34 of AdminBannerForm.tsx, which calls `convertDatetimeStringToNumber()`, which returns either a `number` or `undefined`.

So, why is there a need to check if startDate is an array? Can't we skip the `actualStartDate` variable entirely here, and just test against `startDate` below?

I may have misunderstood here - feel free to correct me. If I haven't misunderstood, then the same change should be made here as in the yup schema on the API side.

benmartin-coforma (commented):
This is really solid work! I'm leaving this review as a Comment, because although I do have outstanding questions, I don't want to Request Changes and be a blocker for my whole vacation.

Thanks!

services/ui-src/src/utils/other/time.test.ts
rocio-desantiago: I am making the executive decision to keep this one as is 😆

services/ui-src/src/utils/other/time.test.ts
rocio-desantiago: `.toBeFalsy()` it is

services/ui-src/src/utils/validation/bannerValidation.ts
rocio-desantiago: I was trying a bunch of weird stuff and I rewrote this part a few times before it worked, so that array stuff is just a vestige of a previous iteration. Good catch!

services/ui-src/src/utils/validation/bannerValidation.ts
```diff
@@ -1,59 +1,103 @@
...
+      const { bannerStartDate, bannerEndDate } = value || {};
+      const startDateString = bannerStartDate?.answer;
+      const endDateString = bannerEndDate?.answer;
+
+      if (startDateString && endDateString) {
+        const startDate = parseMMDDYYYY(startDateString);
+        const endDate = parseMMDDYYYY(endDateString);
+
+        if (startDate && endDate) {
+          if (endDate <= startDate) {
```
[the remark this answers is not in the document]
rocio-desantiago: Oooh that's a good question! I wasn't sure if that was an expected pattern. I'll confirm if so.

benmartin-coforma (approved):

ailZhou (approved):
lgtm!

PR #271, opened 2025-05-21
Extract LengthOfStay from a kind of PerformanceRate to a standalone Element

benmartin-coforma:
### Description
Everything we render within a report page has an ElementType. The HeaderElement, the TextFieldElement, the RadioElement. Most of these are very specific, and correspond to a single "thing" on the page. PerformanceRate is an exception in both respects. It is non-specific, in that there are several kinds of PerformanceRate, and each of those renders multiple related inputs to the page.

This PR takes one of the kinds of PerformanceRate, and elevates it to its own ElementType. It still contains multiple inputs (eight of them, in fact), but since they are very tightly related inputs, it is not inappropriate to do so.

TODO Ben: describe the changes in more detail.

### Related ticket(s)
CMDCT-4450

---
### Important updates

---
### Author checklist

### DIFF
--- a/services/app-api/forms/2026/elements.ts
+++ b/services/app-api/forms/2026/elements.ts
@@ -4,6 +4,7 @@ import {
   DividerTemplate,
   ElementType,
   HeaderTemplate,
+  LengthOfStayRateTemplate,
   MeasureDetailsTemplate,
   MeasureFooterTemplate,
   MeasureResultsNavigationTableTemplate,
@@ -16,6 +17,8 @@ import {
   SubHeaderTemplate,
   TextAreaBoxTemplate,
 } from "../../types/reports";
+
+const REPORT_YEAR = 2026;
 
 export const returnToRequiredDashboard: ButtonLinkTemplate = {
   type: ElementType.ButtonLink,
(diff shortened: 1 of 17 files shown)


services/database/scripts/utils.ts
benmartin-coforma, line 1: I originally wrote this script to support a migration for the PerformanceRate components in existing reports. I ended up deleting that migration, but these utils are likely to come in handy eventually.

benmartin-coforma, line 1: I originally wrote this to support the migration for existing reports' PerformanceRate fields. That migration has been deleted, but I believe these utils will come in handy eventually regardless - constructing clients and iterating scan results is extremely likely functionality for future one-off ETL scripts.

services/ui-src/src/components/rates/calculations.tsx
```diff
@@ -1,62 +1,62 @@
-import { AnyObject } from "yup";
```
benmartin-coforma, line 1: This was the last reference to `AnyObject` in the entire App 😌

services/ui-src/src/components/rates/calculations.tsx
```diff
@@ -1,62 +1,62 @@
...
+ * and we only display `0`, it may seem that the calculation errored out.
+ * Whereas displaying `0.00` gives a nicer impression.
+ */
+export const stringifyResult = (value: number | undefined) => {
+  if (value === undefined) return "";
+  if (value === 0) return (0).toFixed(DECIMAL_PLACES);
+  return value.toString();
 };
 
-export const FacilityLengthOfStayCalc = (
```
benmartin-coforma, line 17: This function has been moved to the only component that uses it, Fields.tsx

services/ui-src/src/components/rates/calculations.tsx
```diff
@@ -1,62 +1,62 @@
...
-  )
-    rate["rar-min-stay"] = roundTo(
-      (rate["opr-min-stay"] / rate["epr-min-stay"]) * rate["multi-plan"],
-      2
-    ).toFixed(2);
-
-  return rate;
-};
-
-export const NDRCalc = (rate: AnyObject, multiplier: number) => {
```
benmartin-coforma, line 56: This function has been inlined (and thus duplicated) within each of the three components that call it. You'll find the `numerator / denominator` embedded within `computeAnswer` in NDR.tsx, NDREnhanced.tsx, and NDRFields.tsx.

services/ui-src/src/components/rates/calculations.tsx
benmartin-coforma, line 1: Note that `stringifyInput` and `stringifyResult` are used for _slightly_ different purposes. `stringifyInput` is used for performanceTargets, numerators, denominators, expectedCounts, actualCounts - values that were typed in by the user. `stringifyResult` is used for rates, actualRates, adjustedRates - values that were computed by us on behalf of the user.

services/ui-src/src/components/rates/types/NDRFields.tsx
```diff
@@ -2,157 +2,215 @@ import React, { useEffect, useState } from "react";
...
+              {fields.map((field, fieldIndex) => {
+                const rateObject = rateSet.rates[fieldIndex];
+
+                return (
+                  <Stack key={`${assess.id}.${field.id}`} gap="2rem">
+                    <Heading variant="nestedHeading">{field.label}</Heading>
+                    <CmsdsTextField
+                      label={labelTemplate
+                        .replace("{{field}}", field.label.toLowerCase())
+                        .replace("{{assessment}}", assess.label)}
```
benmartin-coforma, line 170: Hey look, it's that dynamic text substitution we talked about back in forms/2026/elements.ts!

services/ui-src/src/components/report/Page.tsx
benmartin-coforma, line 1: I reworked this file again because React really doesn't like when you pass the `key` inside of a spread props object:
```tsx
// Boo, a console error, React can't handle this
const props = { foo: "foo", bar: "bar", key: "123" };
return <MyElement {...props}/>;

// React is fine with this though
const props = { foo: "foo", bar: "bar" };
return <MyElement key="123" {...props}/>
```

Applying only that fix, we would get (for example)
```tsx
const props = {
  formkey: // etc
  disabled: // etc
};
// ...
return <HeaderElement key={index} {...{...props, element }} />;
```

Which just doesn't feel that nice to me. I already didn't like having six dots in a row before `props`, and now it's only holding two values? I'd rather skip the second rest and include those values directly:
```tsx
return <HeaderElement key={index} {...{ formkey, disabled, element }} />;
```

One more tweak to reduce the amount of code is to move the `key={index}` out to the calling function, since it has nothing to do with the PageElement we're trying to render. Applying that, and we get the code you see below.

services/ui-src/src/types/other.ts
```diff
@@ -25,11 +21,6 @@ export interface TableContentShape {
   footRow?: string[][];
 }
 
-// OTHER
-export interface AnyObject {
-  [key: string]: any;
-}
```
benmartin-coforma, line 31: 🫡

services/ui-src/src/components/rates/types/NDRFields.tsx
```diff
@@ -2,157 +2,215 @@ import React, { useEffect, useState } from "react";
...
   );
 };
+
+const sx = {
+  performance: {
+    input: {
+      width: "240px",
+    },
+  },
+};
```
benmartin-coforma, line 216: This is the fourth time you've scrolled past this little block of `sx` with the input width - because I copy-pasted it from PerformanceRate.tsx into all four rate components. It feels like maybe this belongs... elsewhere. Higher up in the app. Maybe, much higher? Do we ever want text inputs to be a width other than 240px?

services/ui-src/src/utils/other/arrays.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @example
+ * const a = [1, 2, 3];
+ * const b = ["one", "two", "three"];
+ * zip(a, b) // => [[1, "one"], [2, "two"], [3, "three"]]
+ */
+export const zip = <T, U>(arrayA: T[], arrayB: U[]): [T, U][] => {
+  if (arrayA.length !== arrayB.length) {
+    throw new Error(
+      `Cannot zip arrays of different lengths! Got lengths ${arrayA.length}, ${arrayB.length}`
+    );
```
benmartin-coforma, line 14: ┏┓
┃┃╱╲ In this 
┃╱╱╲╲ house 
╱╱╭╮╲╲ we
▔▏┗┛▕▔ zip
╱▔▔▔▔▔▔▔▔▔▔╲ 
&nbsp;&nbsp;&nbsp;&nbsp;same-length arrays
╱╱┏┳┓╭╮┏┳┓ ╲╲ 
▔▏┗┻┛┃┃┗┻┛▕▔

services/ui-src/src/utils/testing/setupJest.tsx
```diff
@@ -24,6 +24,8 @@ import { mockBannerData } from "./mockBanner";
 
 global.React = React;
 
+global.structuredClone = (val: any) => JSON.parse(JSON.stringify(val));
```
benmartin-coforma, line 27: This is a close-enough implementation of `structuredClone` to use in our unit tests, since it is not provided globally by Jest or JSDom or whoever we need to blame about it.

Note that structuredClone in the real world does _not_ behave like JSON stringify + parse, in a few meaningful ways - but none of them currently come up in our app. Nor, in my opinion, are they likely to.

Filtos (approved):
A few comments, but nothing really blocking. Tested manually and all seems to work properly.

services/ui-src/src/components/rates/calculations.tsx
```diff
@@ -1,62 +1,62 @@
...
+ * Round the given number to {@link DECIMAL_PLACES}.
+ * If it is undefined, return undefined.
+ */
+export const roundRate = (value: number | undefined): number | undefined => {
+  if (value === undefined) return undefined;
+
+  const shift = Math.pow(10, DECIMAL_PLACES);
+  const result = Math.round(value * shift) / shift;
+  if (Object.is(result, -0)) return 0;
+  return result;
```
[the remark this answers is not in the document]
benmartin-coforma: Excellent find. I think "the epsilon method" outlined in one of those answers is the best fit for us here.

services/app-api/forms/2026/elements.ts
```diff
@@ -444,17 +421,17 @@ export const performanceRateTermStay: PerformanceRateTemplate = {
     { id: "med-term", label: "Medium Term Stay" },
     { id: "long-term", label: "Long Term Stay" },
   ],
-  rateType: PerformanceRateType.NDR_FIELDS,
   required: true,
   multiplier: 1000,
```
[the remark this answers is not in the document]
benmartin-coforma: That's exactly what I meant, thank you 😅

services/app-api/forms/2026/elements.ts
```diff
@@ -344,95 +353,63 @@ export const exclusionRatesPersonPlanElements: PerformanceRateTemplate = {
...
-      id: "rar-min-stay",
-      label: "Risk Adjusted Rate for Minimizing Length of Facility Stay",
-      autoCalc: true,
-    },
-  ],
-  rateType: PerformanceRateType.FIELDS,
+  labels: {
+    performanceTarget: `What is the ${
+      REPORT_YEAR + 2
+    } state performance target for this assessment?`,
```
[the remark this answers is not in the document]
benmartin-coforma: Great idea. It hadn't even occurred to me, because I wrote this component first. But I'll do exactly this.

ajaitasaini (approved):

Filtos (approved):

services/ui-src/src/components/rates/types/NDRFields.tsx
rocio-desantiago: Question for Megan?

PR #278, opened 2025-05-30
QMS - UAT - Performance Rate Precision

ajaitasaini:
### Description

For all rates, the displayed value should show up to two numbers beyond the decimal point, and any number before the decimal point.

### Related ticket(s)

CMDCT-4678

---
### How to test

cloudfront link: https://d3w587nfq1mc3p.cloudfront.net/
Confirm that all performance rates round to 2 decimal points (confirm that 0 also rounds to 0.00)
Screenshot 2025-05-30 at 11 11 52 AM

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/rates/calculations.tsx
+++ b/services/ui-src/src/components/rates/calculations.tsx
@@ -50,7 +50,7 @@ export const FacilityLengthOfStayCalc = (
       2
     );
 
-  return rate;
+  return rate.toFixed(2);
 };
 
 export const NDRCalc = (rate: AnyObject, multiplier: number) => {
@@ -58,5 +58,5 @@ export const NDRCalc = (rate: AnyObject, multiplier: number) => {
 
   if (!isFilled(numerator) || !isFilled(denominator)) return "";
 
-  return roundTo(numerator / denominator, 2) * multiplier;
+  return (roundTo(numerator / denominator, 2) * multiplier).toFixed(2);
 };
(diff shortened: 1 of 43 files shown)


rocio-desantiago:
I'm getting NaN when I input 0 to try to get 0.00.
Should a value like 0.5 also show up as 0.50? I'm getting 0.5 when I put 5 for numerator and 10 for denominator for example.

services/ui-src/src/components/rates/calculations.tsx
```diff
@@ -58,5 +58,5 @@ export const NDRCalc = (rate: AnyObject, multiplier: number) => {
 
   if (!isFilled(numerator) || !isFilled(denominator)) return "";
 
-  return roundTo(numerator / denominator, 1) * multiplier;
+  return (roundTo(numerator / denominator, 2) * multiplier).toFixed(2);
```
benmartin-coforma, line 61: I think this is worth getting clarity on. We definitely want to change the second argument to `roundTo` from `1` to `2`, but I don't think we want the `toFixed()` call, and we may also want to apply the multiplier _before_ rounding.

Suppose we're looking at LTSS-6 (which has a multiplier of `1000`), and we have a numerator of `1` and a denominator of `3`. If we round before multiplying, we get a rate of `330`. If we round after multiplying, we get a rate of `333.33`. Which of those is correct?

ajaitasaini: I believe we do need the toFixed() call, as the request is to have whole numbers like 1 be presented as 1.00. We do need to round after multiplying because we would expect a rate of 333.33.

benmartin-coforma (approved):

BearHanded (approved):

PR #292, opened 2025-06-17
Fix some console errors

benmartin-coforma:
### Description
This PR has a few minor changes. They're all inconsequential from an application behavior standpoint, but they each prevent an error in the browser console.

### Related ticket(s)
CMDCT-4508

---
### How to test
https://d38o01ijro6n92.cloudfront.net/

1. Open up your browser's console
2. Click around the app
3. You should see fewer errors than before
   - There should be no "validate DOM nesting" errors
   - There should be no "each child needs a unique key" errors
   - You will still see a bunch of "support for defaultProps will be removed" errors, but those are Not Our Fault. We'll only be able to fix them by updating our version of `@cmsgov/design-system`, once they complete [their fix](https://github.com/CMSgov/design-system/pull/3433/files#diff-29e1b7c17921da17acbfbdf347035e1e8090a6f037da2d3641a463c51a21c688)

### Important updates
n/a

---
### Author checklist

### DIFF
--- a/services/ui-src/src/components/fields/RadioField.tsx
+++ b/services/ui-src/src/components/fields/RadioField.tsx
@@ -32,7 +32,7 @@ const formatChoices = (
     }));
 
     const checkedChildren = [
-      <Box sx={sx.children}>
+      <Box key="radio-sub-page" sx={sx.children}>
         <Page elements={children} />
       </Box>,
     ];
--- a/services/ui-src/src/components/fields/TextAreaField.tsx
+++ b/services/ui-src/src/components/fields/TextAreaField.tsx
@@ -39,10 +39,12 @@ export const TextAreaField = (props: PageElementProps<TextAreaBoxTemplate>) => {
   const errorMessage: string | undefined = get(formErrorState, key)?.message;
   const parsedHint = textbox.helperText && parseHtml(textbox.helperText);
 
-  const labelText = [
-    textbox.label,
-    !textbox.required && <span className="optionalText"> (optional)</span>,
-  ];
+  const labelText = (
+    <>
+      {textbox.label}
+      {!textbox.required && <span className="optionalText"> (optional)</span>}
+    </>
+  );
 
   if (hideElement) {
     return null;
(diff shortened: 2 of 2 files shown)


services/ui-src/src/components/fields/RadioField.tsx
```diff
@@ -32,7 +32,7 @@ const formatChoices = (
     }));
 
     const checkedChildren = [
-      <Box sx={sx.children}>
+      <Box key="radio-sub-page" sx={sx.children}>
```
benmartin-coforma, line 35: Since `checkedChildren` is an array containing JSX, which will eventually be passed to React, every element in the array requires a key. Yes, all one elements.

services/ui-src/src/components/fields/TextAreaField.tsx
```diff
@@ -39,10 +39,12 @@ export const TextAreaField = (props: PageElementProps<TextAreaBoxTemplate>) => {
...
-  const labelText = [
-    textbox.label,
-    !textbox.required && <span className="optionalText"> (optional)</span>,
-  ];
+  const labelText = (
+    <>
+      {textbox.label}
+      {!textbox.required && <span className="optionalText"> (optional)</span>}
+    </>
+  );
```
benmartin-coforma, line 47: Here we had an array containing JSX, so React was complaining that we hadn't specified keys. Rather than adding them, I turned the array into a single JSX object instead.

services/ui-src/src/components/fields/TextAreaField.tsx
```diff
@@ -53,7 +55,7 @@ export const TextAreaField = (props: PageElementProps<TextAreaBoxTemplate>) => {
       <CmsdsTextField
         id={key}
         name={key}
-        label={labelText || ""}
+        label={labelText}
```
benmartin-coforma, line 58: (and that JSX object is always truthy, so there will be no need to default it)

BearHanded (approved):

Filtos (approved):

PR #299, opened 2025-06-25
in radio modal case qmDeliveryMethodChange, update state again

Filtos:
### Description

- Calling to update zustand state again with proper data in case of `qmDeliveryMethodChange`
More detail: While `setAnswers()` is already being called on every change inside `ReportPageWrapper.tsx`, because we are showing a modal to the user before we actually make the switch, the `setAnswers()` is being called without any change. Now in these cases we will call `setAnswers()` again from inside `RadioField.tsx` with the 

### Related ticket(s)

CMDCT-4799

---
### How to test

- Login: https://d2uf3a1nswxoll.cloudfront.net/
- Navigate to LTSS-1
- For Measure Delivery Methods, select an initial radio selection, and refresh to make sure it saved
- Select a different radio selection, confirm on the modal, then refresh. Make sure the radio selection persists

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/fields/RadioField.tsx
+++ b/services/ui-src/src/components/fields/RadioField.tsx
@@ -152,6 +152,7 @@ export const RadioField = (props: PageElementProps<RadioTemplate>) => {
         changeDeliveryMethods(currentPageId, value);
         // Update zustand state with latest form values
         setAnswers(form.getValues());
+        autosave();
         return; // after the clear, allow normal setting of this page to occur
     }
   };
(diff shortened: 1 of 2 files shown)


ailZhou (approved):
lgtms, this looks to fix the issue the saving issue listed in the jira ticket. 

there can still be data lost if we refresh too fast but that's due to the speed of when autosave actually runs.

services/ui-src/src/components/fields/RadioField.tsx
```diff
@@ -149,6 +150,8 @@ export const RadioField = (props: PageElementProps<RadioTemplate>) => {
         return;
       case "qmDeliveryMethodChange":
         changeDeliveryMethods(currentPageId, value);
+        // Update zustand state with latest form values
+        setAnswers(form.getValues());
```
BearHanded, line 154: Does autosave still run after this occurs? That's the other part that is wrapped into the PageWrapper

Filtos: Good catch, during testing I was going through the flow pretty fast and was able update the answers before the `ReportPageWrapper.tsx` autosave was called. Fixed.

PR #302, opened 2025-06-27
Unit test cleanup

benmartin-coforma:
### Description
There are no functionality changes in here - just a whole bunch of housekeeping for our unit tests.
- "`ReactDOMTestUtils.act` is deprecated in favor of `React.act`"
   + We were using an old version of testing-library, which was invoking a deprecated version of `act()`. It still works (for now), but it was complaining a lot in the console.
- "The current testing environment is not configured to support act"
   + We were using `act()` in a whole bunch of places where it wasn't needed. I'm still not sure exactly _why_ those uses were causing this error, since our testing environment definitely supports `act()`, but removing the calls avoids the errors.
- "React state updates should be wrapped into act"
   + A lot of these were resolved by invoking `testA11yAct` instead of `testA11y`. We don't _usually_ need to call `act()` ourselves, because most of our test actions that can cause react state updates are done through `testing-library`, which handles all that for us. But `jest-axe` does not.
      * I'm not sure exactly what `jest-axe` is doing internally that's causing state updates to our components, but if wrapping it in `act()` makes the console errors go away, then I'm not worried about ti.
      * I'm not sure if it makes sense to have both `testA11y` and `testA11yAct`; having both of them with no clear way to know which one to call is a bit of a footgun. It may be a good idea to just _always_ call wrap our `jest-axe` calls in `act()`.
   + The sneakiest of these were for AddEditReportModal.test.tsx. Previously, line 140 was missing an `await`, which was causing this error _only when running all tests at once_. My hair is a little bit grayer for the time & stress spent getting to the bottom of that one. I wish we had some kind of linting rule for non-awaited async calls. Is that a thing?
- "Support for defaultProps will be removed"
   + This is a known source of console errors which is a complete non-issue for us. As soon as CMSDS publishes a version for React 18, we will update it and these error messages will go away.
   + In the meantime, I got so fed up with them that I patched out the defaultProps usage in my own node_modules folder.
   + Then I updated node_modules and my patch got wiped out, of course
   + So I wrote a script that will re-apply the patch. It doesn't run automatically, but you can run it with `yarn patch`. And I recommend you do so.
   + Let me say that again: 🚨🚨 ***Once this is merged, you should run `yarn patch`*** 🚨🚨. You will see a lot less distracting noise in your browser console and unit test output.
- "Each child in a list should have a unique key prop"
   + This wasn't specific to unit tests; I should have caught it in #292. The culprit was the array of elements in StatusAlert.tsx.
- "timers APIs are not replaced with fake timers"
   + Simple enough: Timeout.test.tsx was calling `jest.runAllTimers()` without first having called `jest.useFakeTimers()`. The test was passing without fake timers. So instead of setting them up properly, I just removed the call to run them.
- "500" and "Request Failed - my/url - undefined"
   + This was entirely expected output, from a test that was deliberately exercising a `catch` block. I modified apiLib.test.ts to swallow these logs.
- I also consolidated all of our use of `jest-axe` to go through either `testA11y` or `testA11yAct`. We no longer import that package in any file except `commonTests.tsx`.
- I also replaced a few uses of `fireEvent` with the friendlier `userEvent`. I wish I could remove all of them, but there are a few tricky ones left, which I'm going to call out of scope for this PR.

### Related ticket(s)
CMDCT-4509

---
### How to test
* All the unit tests should pass
* The only output from the unit tests should be the pass/fail indicators
* You can verify this by running `yarn test > temp.txt`, and verifying that temp.txt contains no errors, warnings, or other output. Well, like three lines from yarn itself, but nothing else.

### Important updates
A new patch script to make your browser console quieter! Run it!

---
### Author checklist


services/ui-src/src/components/fields/RadioField.test.tsx
```diff
@@ -149,11 +150,9 @@ describe("Radio field click action logic", () => {
       </div>
     );
     render(deliveryRadio);
-    const radioField = screen.queryByText("Choice 1");
+    const radioField = screen.getByText("Choice 1");
```
benmartin-coforma, line 153: Replacing `query` with `get` here allowed me to remove the `!` from `userEvent.click(radioButton)`. The same for line 230 in this file.

Filtos (approved):
Test passing locally after running `yarn install`.
Not getting any console errors locally after running `yarn patch`.

ailZhou (approved):

PR #312, opened 2025-07-08
[CDMCT-4831] Create PDF Table Components

ailZhou:
### Description

This PR is for adding the Table components into the PDF, it also kickstarts the structure we want to use when rendering new components on the page.

https://github.com/user-attachments/assets/dfc470a6-31fb-4858-b079-ad4f235fbd41

**So What Are These Files?**
1) ExportedReportElements.tsx - Is similar to Page.tsx and where you would add renders for components that aren't part of the table or needs a unique table like performance measure components.
2) ExportedReportTable.tsx - This is the Indicator / Response table made reusable.

**Question for Y'all**
Would folks prefer to have export renders in the same file as the component being rendered? Example:
Screenshot 2025-07-11 at 11 33 08 AM

or to have it all in the ExportedReportElements.tsx file itself
Screenshot 2025-07-11 at 11 36 30 AM

**Notes:** 
1) I didn't read the ticket at first and thought I was suppose to be making all the components and that is why there is code rendering NDREnchanced Elements. I'm leaving it in because it works and could be used as a template for the other NDR sets.
2) In the same vein, I adjusted the paddings of the components to match what is in the figma. Hopefully when adding new elements, there will be less need to adjust spacing.
3) The **chunkBy** function was kindly given to me by @benmartin-coforma 

### Related ticket(s)

CMDCT-4831
CMDCT-4834
CMDCT-4835

---
### How to test

Deploy Link: https://d3rxpbvo1u1u8l.cloudfront.net/

2) Create a new QMS report
3) Click the [**Edit**] button on the dashboard to enter the report
4) You can choose to fill or not fill out the report data before opening the PDF 
5) Go to the **Review & Submit** page and click [**Review PDF**]
6) Check that the PDF is displaying table components

### Important updates

N/A

---
### Author checklist

---

### DIFF
--- a/services/ui-src/src/components/pages/Export/ExportedReportPage.tsx
+++ b/services/ui-src/src/components/pages/Export/ExportedReportPage.tsx
@@ -20,6 +20,7 @@ import {
   MeasurePageTemplate,
   ParentPageTemplate,
   Report,
+  ReportType,
   ReviewSubmitTemplate,
 } from "types";
 import { ExportedReportBanner, ExportedReportWrapper } from "components";
@@ -103,6 +104,7 @@ export const reportDetails = (report: Report) => {
 };
 
 export const reportSubmissionSetUp = (report: Report) => {
+  if (report.type !== ReportType.QMS) return;
   const rows = [
     {
       indicator: "Is your state reporting on the HCBS CAHPS Survey?",
--- a/services/ui-src/src/components/pages/Export/ExportedReportPage.tsx
+++ b/services/ui-src/src/components/pages/Export/ExportedReportPage.tsx
@@ -81,7 +81,7 @@ export const reportTitle = (report: Report) => {
 
 export const reportDetails = (report: Report) => {
   return (
-    <Table>
+    <Table variant={"reportDetails"}>
       <Thead>
         <Tr>
           <Th>Reporting year</Th>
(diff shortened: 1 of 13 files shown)


ajaitasaini (dismissed):
Code looks good to me, and I am seeing the tables in the pdf!

R/e your question, I think I personally like the export renders in the same file as the element itself, but I don't know if there might be a better reason for the alternative

services/ui-src/src/components/pages/Export/ExportedReportPage.tsx
```diff
@@ -69,7 +74,62 @@ export const ExportedReportPage = () => {
...
+    },
+    {
+      indicator: "Is your state reporting on the NCI-AD Survey?",
+      response: report.options.nciad ? "Yes" : "No",
+    },
+    {
+      indicator: "Is your state reporting on the POM Survey?",
+      response: report.options.pom ? "Yes" : "No",
+    },
+  ];
```
Filtos, line 123: Correct me if Im wrong, but I believe these submission questions only apply for the QMS report, so we dont need to render this for the other reports.

Filtos (approved):
Looks good! Left one comment that could potentially be fixed.
As for your question, I do like keeping things together, buuut I feel like the file will get really big and unruly pretty quick, especially if we keep adding unique NDR elements.
Good either way though.

ajaitasaini (approved):

PR #319, opened 2025-07-25
Remove react-hook-form from AddEditReportModal and AdminBannerForm

benmartin-coforma:
### Description
Our app contains five forms:
1. The login page
2. The admin dashboard (allowing them to select a state and report type)
3. The admin banner form (allowing them to create announcement banners)
4. The add/edit report modal (allowing state users to create & rename reports)
5. Every report page (they all share the same code)

The report forms are very complicated, and integrate with React Hook Form for binding & validation. The login page and admin dashboard forms are very simple, and use simple input components directly (from chakra and CMSDS, respectively). The AddEditReportModal and AdminBannerForm are in the middle ground: They contain enough complexity that, when we implemented them, we leveraged our report input components to manage their state & validation.

I would like to rework our report input components not to use RHF, but that means changing AddEditReportModal and AdminBannerForm as well. Rather than adjusting these forms to interact nicely with the reworked report inputs, this PR takes them the other direction: they will now rely directly on CMSDS input components, and our report input components will only be used in reports. So, this PR unblocks that work.

It would have been possible, in theory, to have AddEditReportModal and AdminBannerForm use CMSDS input components _with_ RHF. I had a lot of trouble trying to make that work. My difficulty may be partially explained by my own ineptitude, but I believe some blame is to be shared with RHF itself. If I have trouble making a library do what it was built to do, then I would prefer to avoid using that library. Therefore, this PR removes RHF entirely from these two forms.

Incidentally, this PR also removes `yup` from the AdminBannerForm, where previously it was integrated with RHF. I'm not as strongly opposed to yup as I am to RHF. But given the choice between writing the needed validation directly in place, versus figuring out how to incorporate a validation library, I went with the direct code. This is slightly more verbose for some conditions - notably, the `start date <= end date` condition is duplicated between two handlers now. But it is _also_ easier to understand why & how the code is doing what it's doing (for me, at least - but of course I'm biased, because I wrote it 😅). I am absolutely open to suggestions on how it might be improved.

This PR description is quite long enough, so I'll put additional comments on the diffs.

### Related ticket(s)
* Unblocks work related to CMDCT-4858
* Unblocks work related to CMDCT-4856

---
### How to test
https://d3bnqsdyy5rbbj.cloudfront.net/

1. Test AddEditReportModal as a state user
   * Can you still create new reports, of each type?
   * Can you still edit report names?
   * Is the AddEditReportModal doing anything weird?
2. Test AdminBannerForm as an admin user
   * Can you still create a banner?
   * Is the form doing anything weird?

### Important updates
n/a

---
### Author checklist


services/ui-src/src/components/banners/PreviewBanner.tsx
benmartin-coforma, line 1: The PreviewBanner was only used from the AdminBannerForm. Its only function was to pull banner data out of the form (using RHF), to pass it to a Banner at the bottom of the page. That Banner is now included directly from AdminBannerForm (now at lines 217-223)

services/ui-src/src/components/modals/AddEditReportModal.tsx
benmartin-coforma, line 1: AddEditReportModal has two major differences from the AdminBannerForm. For one, it may be used to edit existing data; the AdminBannerForm is only used to _create_ banners. For two, it has more or fewer inputs depending on the selected report type.

The form's "mode" (so to speak) is determined by `selectedReport`. If it's defined, we're in "edit mode", otherwise we're in "add mode". This may affect the initial state, and it may cause some fields to be readonly, but otherwise has little effect on the form's behavior.

The varying number of inputs is a lot trickier to manage. We always have two inputs for the report name & year, but the remaining inputs (which are collectively called "options") are stashed away in other files. I wanted to keep this pattern, so that this file's size won't depend on the number of report types in the app. But I did rework it significantly. Whereas previously each report type had to provide four things (a verbiage object, a parsing function, an initial state, and a JSX element), now they provide only two (the same verbiage object, and a fully-fledged component, which contains its own parsing & state management). These requirements are defined up-top in the `AddEditReportModalOptions` type, and delivered through the same `buildModalOptions` function as before.

The communication between the parent form (AddEditReportModal) and child component (the report type's OptionsComponent`) is two-way:
* When the `selectedReport` changes, that is passed directly through parent to child.
* When the options data is changed, that is passed from child up to parent through the `onOptionsChange` callback.
* If that change made the options complete (for example, when the user provides a value for the 4th QMS radio button), that fact is passed from child up to parent through the `setOptionsComplete` callback. Until the child has done so, the parent knows that the form cannot be submitted.
* When the form is first loaded, if there is no report selected, all the inputs will be empty but none of them will have errors. If the user clicks submit without filling in the inputs, all the untouched required inputs should show errors. This is made possible by the `submissionAttempted` flag. It is passed from parent to child. The child watches it for changes. When it changes to `true`, the child performs validation on all its inputs at once.

The trickiest thing about this component is the initial state of `isComplete`. For QMS, the options are not complete until the child says so. For TACM and CI, the options are complete by default - because there aren't any! We handle this by looking at the truthiness of `OptionsComponent`. If there is one, then `isComplete` starts at `false` and waits for `setComplete(true)`. If there isn't, then `isComplete` starts true and the form may be submitted as soon as the name and year are provided.

BearHanded (approved):
😎

services/ui-src/src/components/modals/AddFormOptions/QmsOptions.tsx
```diff
@@ -13,128 +15,115 @@ const verbiage = {
...
+                },
+              ]}
+              errorMessage={errorData[key]}
+              onChange={onChange}
+              disabled={!!selectedReport}
+            />
+          );
+        })}
+      </>
+    );
```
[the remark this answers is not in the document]
benmartin-coforma: This is a pretty normal level of complexity for our React components. I suspect Quiltish is subjecting it to a higher standard than usual because the type signature makes it look a little different than our typical component.

Either that, or it's slightly more complex than our typical component, in which case I will retreat from "the bot is broken" to "the bot is just wrong". I think this code is fine.

PR #321, opened 2025-07-29
Remove react-hook-form from report components

benmartin-coforma:
### Description
Well, here we are. This pull request removes `react-hook-form` from HCBS. It also removes `yup` almost entirely from the frontend.

It's not entirely clear if this is a net win in terms of lines of code. This PR removes 1500 lines and adds 1200, but a significant portion of that is in test code, and some of our components gained lines, so it's too close to call. But I do believe it's a win in terms of understandability, for several crucial behaviors in our app:
* Modifying report data. With RHF, this was done through formkeys, which it... somehow... used as a sort of JSON path. Now, it's done through the `updateElement` callback, which you can find and inspect yourself.
* Managing page state. With RHF, this was done through methods like register, reset, and getValues. This led to several bugs (data leaking between pages, form data de-syncing from store data). Now, it's done through props and onChange handlers. Locality of Behavior is increased: More of the code that describes how a component behaves is contained within that component.
* Managing form validation and errors. With RHF, this was done by integrating a dynamic yup model into the form. Therefore, only data that we saved back to the report object could be validated. This required us to choose between saving only valid data, and having different user-facing error messages for data that is invalid in different ways. Now, we can have both. Each component is responsible for its own validation, and only sends valid answers up through its callbacks. It's clear what validation will happen, and when. Again, Locality of Behavior is increased.
* Several complex bits of code disappeared. We no longer need `useLiveElement`, because there is only one source of data: the store, which sends data down through props. We no longer need the dynamic yup schema, as described in the previous bullet. And several higher-level actions are simpler: the store no longer needs to search through report data to find error messages, and the Clear Measure button only needs to talk to the store (and not the form).

The way report pages work, at a high level, is this:
* When the browser visits a report route, the router instantiates a `<ReportPageWrapper>`.
* The ReportPageWrapper reads URL parameters to determine which page it's trying to render, and fetches that page's element list from the store.
* The ReportPageWrapper instantiates a `<Page>`, giving it an array of elements to render and a callback to modify that array. Whenever the Page hits that callback, the ReportPageWrapper will send that data to the store, and call `autosave()`.
* The Page instantiates each element, giving it data through props, and a callback to modify itself. Whenever the element hits that callback, the Page will hit its own callback to send the update up to ReportPageWrapper.
* The element manages three pieces of state:
   + The data it received through its props, including the `answer` property. When that changes, the element updates its display.
   + The data it's currently displaying, usually called `displayValue`. This may be a single string (like for TextField), or a complicated object (like for NDREnhanced). It will resemble the answer, but all values will be strings.
   + The errors it's currently displaying. This will be precisely the same shape as the display value.
* Every time an element receives a click, keystroke, or blur event, it will
   1. Instantly update its display value, whether the value is valid or not.
   2. Instantly update its error value, either to an empty string or to a message describing exactly what's wrong.
   3. Instantly push an updated version of the `answer` up through its callback. If the value that just changed is valid, this will contain that value (or a parsed/masked version of that value, as appropriate for numbers/dates). If the value that just changed is not valid, this will be `undefined` (or an object with an undefined value, as appropriate for NDR & friends).
      - If this is a Radio element, this may also update the `checkedChildren` property, showing child elements.

Incidentally, this PR also implements the requests from Cameron regarding performance rates ( https://coforma.slack.com/archives/C06V6EKEL74/p1751295597060699 )
* User-entered values are no longer rounded (until the 14th decimal place).
* We now store a more precise rate value (14 decimal places) than we display (2 decimal places).

### Related ticket(s)
* CMDCT-4858: Validation Errors spike.
   + Now that we are managing our validation directly, without RHF or Yup, we can implement as many or as few validation behaviors as we like.
* CMDCT-4856: Data Transfer between measures.
   + The root cause of this bug was never clear; possibly we blame RHF, possibly we blame React itself, possibly both. RHF is now out of the picture, and elements on different pages now have different keys (since the page ID is included in each element's key). So this is fixed now.

---
### How to test
Oh, boy. Well, I changed every single report component in the app, as well as some higher-level stuff affecting how reports are saved. So a lot of testing is needed.

https://d5uwjhfxu2z5l.cloudfront.net/

We should verify
* Every input type: Radio, Dropdown, Text, Text Area, Number, Date
   + Except that we can't really test Date fields, because we don't have any date fields in any reports. The only place this component was used was in the AdminBannerForm, which now uses the underlying CMSDS component directly
   + Note that for Number, we need to go into TACM, one of the HAPCH pages, and play around with the Sample Size field. QMS has no Number inputs.
   + Note that Radio has special behavior for
      - Showing/hiding child questions
      - Showing/hiding questions elsewhere on the page (the reporting radio button). Hiding these questions should also wipe their answers.
      - Enabling/disabling rows of the Measure Results Navigation Table. Again, disabling a delivery method should wipe all answers within that delivery method
* Every auto-calculation component: NDR, NDREnhanced, NDRFields, NDRBasic, and Fields
   + Note that for NDRBasic, we again need to go into TACM
* Report data is saved, and can be loaded again (for example, when the page is refreshed)
* Measures can be cleared, which should also reset their detail pages
* Data does not bleed between HAPCH-1 and HAPCH-2

Am I missing anything?

And this should absolutely be reviewed by Design and Product - in particular, the way inputs behave:
* on blur: for example, clicking into an empty input and then clicking out again before entering a value
* on invalid input: for example, when a partial date has been entered, or a non-numeric value
* All possible error messages. For example, should we have different error messages for invalid values vs missing values? What about invalid values on a required field vs an optional field?

### Important updates
Farewell, RHF.

---
### Author checklist


rocio-desantiago (approved):
Tested in deploy and successfully submitted each report. 
1. Tiny question about email validation
2. Found the weirdest tabbing bug that a user probably would never follow but probs still good to follow up, when back-tabbing through the complete measure button after already clicking it, disabled becomes false after being set to true because it was "completed"
 
Attaching video

https://github.com/user-attachments/assets/b1d82639-11e7-4673-910d-42e0e0f3d108

Filtos (approved):
Huzzah!

Filtos (approved):

PR #324, opened 2025-07-30
Update HCBS to manage log groups and update CDK constructs

JonHolman:
### Description
Update HCBS to manage log groups and update CDK constructs

### Related ticket(s)

CMDCT-4976

---
### How to test
Open deployed instance and confirm it works https://dp3ckw9mwh6qu.cloudfront.net/

### Important updates

---
### Author checklist

- [X] I have performed a self-review of my code
- [X] I have added [thorough](https://shorturl.at/aejkF) tests, if necessary
- [X] I have updated relevant documentation, if necessary
---

### DIFF
--- a/deployment/constructs/lambda.ts
+++ b/deployment/constructs/lambda.ts
@@ -26,7 +26,6 @@ interface LambdaProps extends Partial<NodejsFunctionProps> {
   stackName: string;
   api?: apigateway.RestApi;
   additionalPolicies?: PolicyStatement[];
-  requestParameters?: string[];
   requestValidator?: apigateway.IRequestValidator;
   isDev: boolean;
 }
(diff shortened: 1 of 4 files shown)


deployment/constructs/lambda.ts
```diff
@@ -94,6 +109,15 @@ export class Lambda extends Construct {
           authorizationType: isLocalStack
             ? undefined
             : apigateway.AuthorizationType.IAM,
+          requestParameters: requestParameters
```
peoplespete, line 112: Not sure what benefit we are really getting from requestParameters at all. Is it important of could it be left out in all apps without a problem?

JonHolman: Discussed and we don't need request parameters in the construct

PR #326, opened 2025-08-01
Prevent no-op "updates" from resetting measure status

benmartin-coforma:
### Description
This should address [the issue Rocio noted](https://github.com/Enterprise-CMCS/macpro-mdct-hcbs/pull/321#pullrequestreview-3073369325) on #321. Specifically: tabbing through a completed measure page, without changing any values, puts the measure back into "In Progress" status.

It was caused because several report components call their `onChange` handler on blur. This resulted in a store action, which assumed that the values had changed, and therefore changed the page status.

I still believe that calling onChange within those components is a good idea; it's a really neat way to make sure that the validation/error message is up-to-date without writing any additional code. Which leaves two options: Either make every component's onChange handler smart enough to not call `updateElement` unless its answer has changed, or check for those changes at a higher level.

I chose to check for changes at the higher level, because it removes the possibility that a future component author will forget/neglect to implement their own dirty check, and be bit by this same bug. This may have performance implications; a deep equality check of the entire page may end up being a lot of work to do on every keystroke. On the other hand, it may not. If it does, one option for optimization would be to move that check into Page.tsx. Then, we would only need a deep equality check of the element in question. If that still ends up being too much work, we may need to revisit my decision and implement those checks within the individual onChange handlers.

### Related ticket(s)
N/A

---
### How to test
https://d1lv5kgr2qqxkx.cloudfront.net/

1. Complete a measure. Change an input. Confirm that the measure is no longer completed.
2. Complete the measure again. Tab through it without changing anything, and confirm that the measure is still completed.
3. Repeat steps 1 and 2 for a measure details page. Confirm that the parent measure's status behaves as expected.

### Important updates
n/a

---
### Author checklist

### DIFF
--- a/services/ui-src/src/utils/state/reportLogic/reportActions.test.ts
+++ b/services/ui-src/src/utils/state/reportLogic/reportActions.test.ts
@@ -22,6 +22,8 @@ import {
   saveReport,
   markPageComplete,
   changeDeliveryMethods,
+  deepEquals,
+  deepMerge,
 } from "./reportActions";
 import {
   mock2MeasureTemplate,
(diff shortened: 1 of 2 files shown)


BearHanded (approved):
very nice testing suite

rocio-desantiago (approved):
Tested and it works ✅

PR #330, opened 2025-08-11
QMS - PDF - Arrange Measures to Match PDF Format

ajaitasaini:
### Description

This PR addresses the rearranging of PDF sections so we can group them by "Required Measures" and "Optional Measures". This required some refactoring of how we render the sections, and involved rearranging the measures and manually inserting the headers before the rendering.

### Related ticket(s)

CMDCT-4838

---
### How to test

https://d83m3pjfgwocz.cloudfront.net/
1. Create a QMS report and fill out whatever measures you'd like (my recommendation is to test without any measures filled out, and then with some required measures + optional measures) 
2. See that the PDF renders correctly and that required measures come before optional measures and that the headers are in the right spot. 
3. If there are no optional measures filled out the section should not show. If there are no required measures filled out the section SHOULD still show

Visually the PDF should match the figma: https://www.figma.com/design/B8IEoYKQ7DuQyx5nStls8F/HCBS?node-id=6078-33747&t=IHupzr9Q2UqMS651-0

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/app-api/forms/2026/elements.ts
+++ b/services/app-api/forms/2026/elements.ts
@@ -208,6 +208,7 @@ export const additionalNotesField: TextAreaBoxTemplate = {
     controllerElementId: "measure-reporting-radio",
     answer: "no",
   },
+  required: false,
 };
 
 export const measureDeliveryMethodsSubheader = [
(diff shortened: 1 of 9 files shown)


benmartin-coforma (commented):
This is solid work, thanks Ajaita!

rocio-desantiago (approved):
Tested each scenario ✅

Filtos (changes requested):
Looks like this PR somehow broke the PDF functionality for CI and TACM reports.
For CI, it only shows a section if it has data inputted (I believe it should always show all sections as they are all required.)
For TACM, all sections after general-info are always missing.

PR #338, opened 2025-08-22
CMDCT-5017

Filtos:
### Description

- Added a mock state to the component inventory
- Added the rest of ElementTypes to the component inventory
- Removed as we were not using it anywhere (we did not even have a component created for it)
- Was running into a circular import issue so had to modify buildComponentDisplay to defer the rendering until runtime

### Related ticket(s)

CMDCT-5017

---
### How to test

- Login: https://dmilsw42ulvex.cloudfront.net/
- Got to `/component-inventory` page and verify that new ElementTypes are rendering

### Important updates

---
### Author checklist

---

### DIFF
--- a/services/app-api/forms/2026/templates.test.ts
+++ b/services/app-api/forms/2026/templates.test.ts
@@ -1,4 +1,4 @@
-import { isResultRowButton, PageType } from "../../types/reports";
+import { PageType } from "../../types/reports";
 import { ciReportTemplate } from "./ci/ci";
 import { CMIT_LIST } from "./cmit";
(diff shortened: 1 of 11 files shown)


Filtos:
@benmartin-coforma Without it in some cases I was getting `Uncaught ReferenceError: Cannot access '<ElementType>' before initialization
    at elementObject.tsx`

rocio-desantiago (commented):
I can't figure out why the first element from the measures element is being cut off like this 🤔 

Image

BearHanded (approved):
I think this looks good. We might want a follow-up ticket for some of the squashing and styling of components on the page, but this tackled the first hard part.

PR #386, opened 2025-10-03
[CMDCT-5154] PDF - Align Styles with designs

ailZhou:
### Related ticket(s)

CMDCT-5154

---

### Notes

---

### Pre-review checklist

---

#### Review

#### Security

---

### DIFF
--- a/services/ui-src/src/components/export/ExportedReportWrapper.tsx
+++ b/services/ui-src/src/components/export/ExportedReportWrapper.tsx
@@ -22,6 +22,10 @@ export const renderReportDisplay = (
   return elements?.map((element: ReportTableType) => element.response);
 };
 
+const isValidHelperText = (helperText: string) => {
+  return !helperText.includes("Warning:") ? helperText : "";
+};
+
 export const ExportedReportWrapper = ({ section }: Props) => {
   const filteredElements = section.elements?.filter((element) => {
     const hasAnswer =
@@ -38,7 +42,10 @@ export const ExportedReportWrapper = ({ section }: Props) => {
     filteredElements?.map((element) => {
       return {
         indicator: "label" in element ? element.label ?? "" : "",
-        helperText: "helperText" in element ? element.helperText : "",
+        helperText:
+          "helperText" in element && element.helperText
+            ? isValidHelperText(element.helperText)
+            : "",
         response: renderElements(section as MeasurePageTemplate, element),
         type: element.type ?? "",
         required: "required" in element ? element.required : false,
--- a/services/ui-src/src/components/export/ExportedReportWrapper.tsx
+++ b/services/ui-src/src/components/export/ExportedReportWrapper.tsx
@@ -41,6 +41,7 @@ export const ExportedReportWrapper = ({ section }: Props) => {
         helperText: "helperText" in element ? element.helperText : "",
         response: renderElements(section as MeasurePageTemplate, element),
         type: element.type ?? "",
+        required: "required" in element ? element.required : false,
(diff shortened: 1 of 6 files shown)


ajaitasaini (approved):

services/ui-src/src/components/export/ExportedReportWrapper.tsx
```diff
@@ -34,13 +40,38 @@ export const ExportedReportWrapper = ({ section }: Props) => {
...
+          .flatMap((choice) => choice?.checkedChildren ?? [])
+      );
+      //Note: answer is be modified from key value to label value from this point onward
+      modifiedElemet.answer = modifiedElemet.choices.find(
+        (choice) => choice.value === modifiedElemet.answer
+      )?.label;
+    }
+
+    expandElements.push(...child);
+  });
```
benmartin-coforma, line 62: * I think this will be a good spot to use `.flatMap()` over repeated `.push()` calls.
* There's no need to clone the element before modifying it; the entire `report.pages` array has already been cloned by this point.
* It might be better to check for `"choices" in element` rather than `element.type === "radio"`, to future-proof against checkbox lists
* I don't think this will work for a radio button nested inside of another radio button? Unless I am wrong, we will need some recursion here.

Put that together, and you get something like this:
```ts
const expandCheckedChildren = (
  elements: PageElement[]
): PageElement[] => {
  return elements.flatMap((element) => {
    if ("choices" in element) {
      const checkedChoice = element.choices.find(
        (choice) => choice.value === element.answer
      );

      // Note that from this point on, the answer is a label and not a key.
      element.answer = checkedChoice?.label;

      // A radio element is immediately followed by all of its child elements.
      return [
        element,
        ...expandCheckedChildren(checkedChoice?.checkedChildren ?? [])
      ]
    } else {
      // All other element types stand on their own.
      return [element];
    }
  });
};

...
const expandedElements = expandCheckedChildren(filteredElements);
```

And here's a little unit test, to confirm that's working as intended:
```ts
  it("should recursively expand checked children", () => {
    const elements: PageElement[] = [
      {
        type: ElementType.Radio,
        id: "banana-preference",
        label: "Do you like bananas?",
        required: true,
        answer: "bananya",
        choices: [
          { label: "No", value: "bananah" },
          {
            label: "Yes, I love bananas",
            value: "bananya",
            checkedChildren: [
              {
                type: ElementType.Radio,
                id: "banana-desire",
                label: "Do you want some bananas?",
                required: true,
                answer: "yes",
                choices: [
                  { label: "I'm good thanks", value: "no" },
                  {
                    label: "Yeah, I'm starving",
                    value: "yes",
                    checkedChildren: [
                      {
                        type: ElementType.TextAreaField,
                        id: "banana-count",
                        label: "Describe your ideal banana",
                        required: true,
                        answer: "Oblong, yellow, not too soft"
                      },
                    ],
                  },
                ],
              },
            ],
          },
        ],
      },
    ];

    render(<ExportedReportWrapper section={{ ...section, elements }} />);

    // All answers should render together in the table
    expect(screen.getAllByRole("table")).toHaveLength(1);
    expect(screen.getByText("Yes, I love bananas")).toBeVisible();
    expect(screen.getByText("Yeah, I'm starving")).toBeVisible();
    expect(screen.getByText("Oblong, yellow, not too soft")).toBeVisible();

    // The choice IDs should not render anywhere
    expect(screen.queryByText("bananah")).not.toBeInTheDocument();
    expect(screen.queryByText("bananya")).not.toBeInTheDocument();
  });
```

PR #495, opened 2025-12-02
DO NOT MERGE - demo application based on HCBS - PR for feedback

JonHolman:
### Description
This is a demo application based on HCBS.  This PR should never be merged.  This PR is for discussions and feedback.

### Related ticket(s)

CMDCT-

---
### Notes

---
### Pre-review checklist

---
#### Review

#### Security


cli/commands/local.ts
```diff
@@ -46,6 +46,8 @@ export const local = {
     process.env.AWS_ACCESS_KEY_ID = "localstack";
     process.env.AWS_SECRET_ACCESS_KEY = "localstack"; // pragma: allowlist secret
     process.env.AWS_ENDPOINT_URL = "http://localhost.localstack.cloud:4566";
+    process.env.AWS_ENDPOINT_URL_S3 =
```
JonHolman: The latest `aws-cdk-local` (I believe v3) requires this to be set as well.

deployment/constructs/dynamodb-table.ts
```diff
@@ -36,7 +38,7 @@ export class DynamoDBTable extends Construct {
       streamable = true,
     } = props;
 
-    const tableName = `${stage}-${name}`;
+    const tableName = `${project}-${stage}-${name}`;
```
JonHolman: this will not be easy to roll out across our apps, but I think it would be nice to have the project name at the beginning of resource names.

deployment/package.json
JonHolman: some restructuring of our package.json files.

deployment/tsconfig.json
JonHolman: some restructuring of our tsconfig.json files

services/app-api/utils/authentication.ts
JonHolman: the latest version of jwt-decode

services/ui-src/src/components/fields/RadioField.tsx
```diff
@@ -5,7 +5,6 @@ import { ChoiceTemplate, PageElement, RadioTemplate } from "types";
 import { parseHtml, useStore } from "utils";
 import { ChoiceList as CmsdsChoiceList } from "@cmsgov/design-system";
 import { Page } from "components/report/Page";
-import { ChoiceProps } from "@cmsgov/design-system/dist/react-components/types/ChoiceList/ChoiceList";
```
JonHolman: I could not find an appropriate type in the latest `@cmsgov/design-system`.  The checks are happy without it.

services/ui-src/src/utils/other/time.ts
JonHolman: functions renamed in the latest `date-fns-tz`

services/ui-src/src/utils/api/providers/ApiProvider.tsx
```diff
@@ -1,13 +1,12 @@
 import { ReactNode, useMemo, createContext } from "react";
 
-export const ApiContext = createContext(null);
```
JonHolman: with this set to null, down below the code errors:

```
src/utils/api/providers/ApiProvider.tsx:11:31 - error TS2322: Type '{}' is not assignable to type 'null'.

11   return <ApiContext.Provider value={values}>{children}</ApiContext.Provider>;
                                 ~~~~~

  node_modules/@types/react/index.d.ts:536:9
    536         value: T;
                ~~~~~
    The expected type comes from property 'value' which is declared here on type 'IntrinsicAttributes & ProviderProps<null>'


Found 1 error in src/utils/api/providers/ApiProvider.tsx:11
```

services/ui-src/package.json
```diff
@@ -97,10 +89,6 @@
       "src"
     ],
     "verbose": true,
-    "moduleNameMapper": {
```
JonHolman: I'm not sure if I should have removed this, but everything seems to work without.

tests/playwright/banner.spec.ts
```diff
@@ -26,9 +26,7 @@ test.describe("Banner functionality", () => {
     await assertBannerIsPopulated(alertPreview, testBannerData);
   });
 
-  test.skip("The banner should be visible from other pages", async ({
```
JonHolman: this test is passing, removed the skip.

deployment/constructs/lambda.ts
```diff
@@ -8,10 +8,10 @@ import { Duration, RemovalPolicy, aws_s3 as s3 } from "aws-cdk-lib";
 import { Runtime } from "aws-cdk-lib/aws-lambda";
 import { PolicyStatement } from "aws-cdk-lib/aws-iam";
 import * as apigateway from "aws-cdk-lib/aws-apigateway";
-import { isLocalStack } from "../local/util";
+import { isLocalStack } from "../local/util.js";
```
gmrabian: nit: since we're setting this up for the first time could we not configure the package and tsconfig so we don't need `.js` extensions on imports?

JonHolman: I wanted to do this, but I couldn't find a way to without configuring tsconfig.json with settings I don't think we should have.  `moduleResolution: bundler` would alllow us to do that, but I don't think that's right.

PR #617, opened 2026-03-17
[HCBS] Add captions to tables

rocio-desantiago:
### Description
- Several `<table>`'s were missing a `<caption>`.
- This PR adds captions to several tables with Chakra-UI: `<TableCaption>` and `<VisuallyHidden>`

Pages that have tables missing captions:
> /report/QMS/`<state>`/`<id>`/req-measure-result
> /report/QMS/`<state>`/`<id>`/LTSS-1
> /report/QMS/`<state>`/`<id>`/LTSS-2
> /report/QMS/`<state>`/`<id>`/LTSS-6
> /report/QMS/`<state>`/`<id>`/LTSS-7
> /report/QMS/`<state>`/`<id>`/LTSS-8
> /report/QMS/`<state>`/`<id>`/POM-1
> /report/QMS/`<state>`/`<id>`/POM-2
> /report/QMS/`<state>`/`<id>`/POM-3
> /report/QMS/`<state>`/`<id>`/POM-4
> /report/QMS/`<state>`/`<id>`/POM-5
> /report/QMS/`<state>`/`<id>`/optional-measure-result
> /report/QMS/`<state>`/`<id>`/FASI-1
> /report/QMS/`<state>`/`<id>`/FASI-2
> /report/QMS/`<state>`/`<id>`/HCBS-10
> /report/QMS/`<state>`/`<id>`/LTSS-3
> /report/QMS/`<state>`/`<id>`/LTSS-4
> /report/QMS/`<state>`/`<id>`/MLTSS-5
> /report/QMS/`<state>`/`<id>`/MLTSS
> /report/QMS/`<state>`/`<id>`/POM-6
> /report/QMS/`<state>`/`<id>`/POM-7
> /report/QMS/`<state>`/`<id>`/review-submit
> /report/QMS/`<state>`/`<id>`/export

#### Solution:

- Add a `<caption>` to tables that meaningfully describe the table content. (Captions can be the same as their nearest heading.)
- Captions are useful for people who use screen readers to navigate directly to tables

### Related ticket(s)

CMDCT-5624

---
### How to test

1. Login as a state user
2. Update Voiceover settings to mute speech to avoid stimulation overload, if needed.

     Voiceover utility menu to select mute speech checkbox
3. Open [deployed environment](https://d1q0y5vrx0094h.cloudfront.net/) in Safari
4. Navigate to tables
    a. Dashboard Table of each report
    b. Required Measure Results table
    c. Optional Measure Results table
    d. Each Measure page per the "Description" (Jira ticket)
    d. Review & Submit table
    e. PDF view (this one has the most tables)   
5. Toggle Voiceover On (This image shows it toggled off - for the record)
     
    Voiceover toggled off

6. Press keys `Ctrl` + `Option `+ `U`. 
7. This will open a menu.  Press right arrow until you see a table menu. 

Best case scenario ⭐ 
This is an example of the tables being defined with more than just the number of columns and rows
PDF view with table captions - rotor view example

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/app-api/utils/reportValidation.ts
+++ b/services/app-api/utils/reportValidation.ts
@@ -274,6 +274,7 @@ const measureTableTemplateSchema = object().shape({
 const eligibilityTableSchema = object().shape({
   type: string().required().matches(new RegExp(ElementType.EligibilityTable)),
   id: string().required(),
+  caption: string().required(),
   fieldLabels: object().shape({
     title: string().required(),
     description: string().required(),
--- a/services/app-api/utils/reportValidation.ts
+++ b/services/app-api/utils/reportValidation.ts
@@ -268,6 +268,7 @@ const measureTableTemplateSchema = object().shape({
   type: string().required().matches(new RegExp(ElementType.MeasureTable)),
   id: string().required(),
   measureDisplay: string().oneOf(["required", "optional"]).required(),
+  caption: string().required(),
 });
 
 const eligibilityTableSchema = object().shape({
@@ -636,6 +637,7 @@ const reportValidateSchema = object().shape({
   year: number().required(),
   submissionCount: number().required(),
   archived: boolean().required(),
+  caption: string().notRequired(),
   options: optionsSchema,
   pages: pagesSchema,
 });
--- a/services/app-api/utils/reportValidation.ts
+++ b/services/app-api/utils/reportValidation.ts
@@ -637,7 +637,6 @@ const reportValidateSchema = object().shape({
   year: number().required(),
   submissionCount: number().required(),
   archived: boolean().required(),
-  caption: string().notRequired(),
   options: optionsSchema,
   pages: pagesSchema,
 });
--- a/services/ui-src/src/components/export/ExportedReportElements.test.tsx
+++ b/services/ui-src/src/components/export/ExportedReportElements.test.tsx
@@ -36,6 +36,9 @@ const section = {
 };
 
 describe("Test ExportedReportElements", () => {
+  beforeEach(() => {
+    section.elements = [];
+  });
   test("Test render SubHeader element", () => {
     const element = renderElements(section, {
       id: "mock-sub-header",
--- a/services/ui-src/src/components/export/ExportedReportElements.test.tsx
+++ b/services/ui-src/src/components/export/ExportedReportElements.test.tsx
@@ -63,10 +63,10 @@ describe("Test ExportedReportElements", () => {
     const element = renderElements(section, mockNDR);
     render(element);
     expect(
-      screen.getByText(
+      screen.getAllByText(
         "Performance Rate : Person uses the same environments as people without disabilities"
-      )
-    ).toBeInTheDocument();
+      ).length
+    ).toBeGreaterThan(0);
     expect(screen.getByRole("row", { name: "Numerator 5" })).toBeVisible();
     expect(screen.getByRole("row", { name: "Denominator 3" })).toBeVisible();
     expect(screen.getByRole("row", { name: /Rate .* 1\.67/ })).toBeVisible();
@@ -76,10 +76,12 @@ describe("Test ExportedReportElements", () => {
     const element = renderElements(section, mockedNDREnhanced);
     render(element);
 
-    expect(screen.getByText("test label: assessment 1")).toBeInTheDocument();
     expect(
-      screen.getByText("Performance Rates Denominator")
-    ).toBeInTheDocument();
+      screen.getAllByText("test label: assessment 1").length
+    ).toBeGreaterThan(0);
+    expect(
+      screen.getAllByText("Performance Rates Denominator").length
+    ).toBeGreaterThan(0);
     expect(screen.getAllByText("2")).toHaveLength(2);
     expect(screen.getAllByText("Not answered")).toHaveLength(1);
   });
--- a/services/ui-src/src/components/export/ExportedReportElements.test.tsx
+++ b/services/ui-src/src/components/export/ExportedReportElements.test.tsx
@@ -66,7 +66,7 @@ describe("Test ExportedReportElements", () => {
       screen.getAllByText(
         "Performance Rate : Person uses the same environments as people without disabilities"
       ).length
-    ).toBeGreaterThan(0);
+    ).toBe(2);
     expect(screen.getByRole("row", { name: "Numerator 5" })).toBeVisible();
     expect(screen.getByRole("row", { name: "Denominator 3" })).toBeVisible();
     expect(screen.getByRole("row", { name: /Rate .* 1\.67/ })).toBeVisible();
--- a/services/ui-src/src/components/export/ExportedReportTable.tsx
+++ b/services/ui-src/src/components/export/ExportedReportTable.tsx
@@ -27,7 +27,7 @@ export type ReportTableType = {
 
 interface Props {
   rows: ReportTableType[];
-  caption?: string;
+  caption: string;
 }
 
 export const ExportedReportTable = ({ rows, caption }: Props) => {
@@ -77,7 +77,7 @@ export const ExportRateTable = (
         <Heading as="h5" variant="h5" className="performance-rate-header">
           {data?.label}
         </Heading>
-        <ExportedReportTable rows={data?.rows} />
+        <ExportedReportTable rows={data?.rows} caption={data.label} />
       </Box>
     )
   );
--- a/services/ui-src/src/components/export/ExportedReportTable.tsx
+++ b/services/ui-src/src/components/export/ExportedReportTable.tsx
@@ -1,14 +1,15 @@
 import {
+  Box,
+  Heading,
   Table,
+  TableCaption,
   Thead,
   Th,
   Tr,
-  Tbody,
   Td,
+  Tbody,
   Text,
-  Heading,
-  Box,
-  TableCaption,
+  VisuallyHidden,
 } from "@chakra-ui/react";
 import { notAnsweredText } from "../../constants";
 import { ElementType } from "types";
@@ -21,6 +22,7 @@ export type ReportTableType = {
   helperText?: string;
   type?: ElementType;
   required?: boolean;
+  caption?: string;
 };
 
 interface Props {
@@ -37,7 +39,9 @@ export const ExportedReportTable = ({ rows, caption }: Props) => {
 
   return (
     <Table variant="export">
-      <TableCaption>{caption}</TableCaption>
+      <TableCaption>
+        <VisuallyHidden>{caption}</VisuallyHidden>
+      </TableCaption>
       <Thead>
         <Tr>
           <Th>Indicator</Th>
--- a/services/ui-src/src/components/export/ExportedReportTable.tsx
+++ b/services/ui-src/src/components/export/ExportedReportTable.tsx
@@ -8,6 +8,7 @@ import {
   Text,
   Heading,
   Box,
+  TableCaption,
 } from "@chakra-ui/react";
 import { notAnsweredText } from "../../constants";
 import { ElementType } from "types";
@@ -24,9 +25,10 @@ export type ReportTableType = {
 
 interface Props {
   rows: ReportTableType[];
+  caption?: string;
 }
 
-export const ExportedReportTable = ({ rows }: Props) => {
+export const ExportedReportTable = ({ rows, caption }: Props) => {
   const getTextColor = (element: ReportTableType) => {
     return element.response === notAnsweredText && element.required
       ? "palette.error_darker"
@@ -35,6 +37,7 @@ export const ExportedReportTable = ({ rows }: Props) => {
 
   return (
     <Table variant="export">
+      <TableCaption>{caption}</TableCaption>
       <Thead>
         <Tr>
           <Th>Indicator</Th>
--- a/services/ui-src/src/components/export/ExportedReportTable.tsx
+++ b/services/ui-src/src/components/export/ExportedReportTable.tsx
@@ -69,7 +69,8 @@ export const ExportedReportTable = ({ rows, caption }: Props) => {
 };
 
 export const ExportRateTable = (
-  tableData: { label: string; rows: ReportTableType[] }[]
+  tableData: { label: string; rows: ReportTableType[] }[],
+  caption?: string
 ) => {
   return tableData.map(
     (data: { label: string; rows: ReportTableType[] }, idx) => (
@@ -77,7 +78,10 @@ export const ExportRateTable = (
         <Heading as="h5" variant="h5" className="performance-rate-header">
           {data?.label}
         </Heading>
-        <ExportedReportTable rows={data?.rows} caption={data.label} />
+        <ExportedReportTable
+          rows={data?.rows}
+          caption={caption || data.label}
+        />
       </Box>
     )
   );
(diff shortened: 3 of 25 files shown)


services/ui-src/src/components/export/ExportedReportTable.tsx
```diff
@@ -35,6 +39,9 @@ export const ExportedReportTable = ({ rows }: Props) => {
 
   return (
     <Table variant="export">
+      <TableCaption>
+        <VisuallyHidden>{caption}</VisuallyHidden>
+      </TableCaption>
```
ntsummers1, line 44: Question: Should we wrap this to check if caption is undefined and if it is, not render this bit? 

Or another thought, should this bit of code stay and instead caption be considered required? I think it makes sense to require it to be passed to the tables to make sure we stay compliant

services/ui-src/src/components/export/ExportedReportWrapper.tsx
```diff
@@ -84,14 +91,39 @@ export const ExportedReportWrapper = ({ section }: Props) => {
...
+        tableGroupActive = false;
+        if (element.type === ElementType.SubHeader && "text" in element) {
+          pendingCaption = buildCaption(element.text as string);
+        }
+      }
+
+      const caption =
+        isTableEl && !tableGroupActive ? pendingCaption : undefined;
+      if (isTableEl) tableGroupActive = true;
+
```
ntsummers1: Nit: I'd def throw in a comment here explaining whats going on. Took me a moment because I was looking into why the `if (isTableEl) tableGroupActive = true` line was there when we were returning on the next line and weren't using it.

rocio-desantiago: Good call 💡

cy-by (dismissed):
I didn't review the code deeply, but I did go through QMS pages with tables using Safari + VoiceOver and Windows + Chrome + JAWS and table captions were correctly being spoken 🙌🏻

services/ui-src/src/components/export/ExportedReportWrapper.tsx
rocio-desantiago: Curious if the comments added are helpful or should I consider JSDoc type comment?

services/ui-src/src/components/export/ExportedReportElements.test.tsx
```diff
@@ -63,10 +63,10 @@ describe("Test ExportedReportElements", () => {
     const element = renderElements(section, mockNDR);
     render(element);
     expect(
-      screen.getByText(
+      screen.getAllByText(
         "Performance Rate : Person uses the same environments as people without disabilities"
-      )
-    ).toBeInTheDocument();
+      ).length
+    ).toBeGreaterThan(0);
```
stacy-coforma: This is the nittiest nit of all but would it make sense to use `toHaveLength(2)` (or whatever length we expect) versus just checking that there is at least 1?  (Obviously this is the first file I am looking at 😅)

services/ui-src/src/components/pages/Export/ExportedReportPage.test.tsx
```diff
@@ -73,9 +73,11 @@ describe("ExportedReportPage", () => {
...
   it("ExportReportPage is visible", () => {
     render(<ExportedReportPage></ExportedReportPage>);
-    expect(
-      screen.getByText("Colorado Quality Measure Set Report for: mock-title")
-    ).toBeInTheDocument();
+    const elements = screen.getAllByText(
+      "Colorado Quality Measure Set Report for: mock-title"
+    );
+    expect(elements.length).toBeGreaterThan(0);
+    elements.forEach((element) => expect(element).toBeInTheDocument());
```
stacy-coforma:
```suggestion
    expect(elements.length).toBeGreaterThan(0);
```

I think checking the length is sufficient here; no need to check each element is in the document since if we have it returned with the `getAllByText` then we can assume it's in the document.  I'd also echo my earlier comment around looking for a specific length if that is feasible.  (More nits.)

services/ui-src/src/components/report/WwlComponents/EligibilityTable.test.tsx
```diff
@@ -141,7 +141,7 @@ describe("<EligibilityTableElement />", () => {
   describe("EligibilityTableElementExport", () => {
     it("should render export table", () => {
       render(EligibilityTableElementExport(mockedElement));
-      expect(screen.getByText("mockTitle1")).toBeVisible();
+      expect(screen.getAllByText("mockTitle1").length).toBeGreaterThan(0);
```
stacy-coforma: This one I wonder if we do still want to check for visibility - at least for one element?  If it is important for at least one instance of this to be visible.  🤔 

Sorry all my comments are nitpicks around tests.

services/ui-src/src/components/report/StatusTable.tsx
```diff
@@ -86,11 +88,13 @@ export const StatusTableElement = () => {
...
     <>
       <Table variant="status">
+        <VisuallyHidden>
+          <TableCaption>Review & Submit</TableCaption>
+        </VisuallyHidden>
         <Thead>
           <Tr>
             <Th>Section</Th>
             <Th>Status</Th>
-            <Th>Actions</Th>
```
stacy-coforma, line 93: Just checking that this was intentional and not the result of a merge conflict? 😅

stacy-coforma (commented):
Overall looks good to me - I just left one comment that I want to confirm before approving 😅 the rest of what I left are incredible nits and I'm happy to go either way.

services/ui-src/src/components/report/StatusTable.tsx
rocio-desantiago: Ahhhhh you're right! ahhhhhh let me undo this. Thanks for noticing!

services/ui-src/src/components/export/ExportedReportTable.tsx
benmartin-coforma: I'm against the conditional. I don't know what it would take to have Typescript enforce this for us, so that may or may not be a good idea, but all tables should have captions.

services/ui-src/src/components/report/MeasureTable.tsx
```diff
@@ -82,6 +84,11 @@ export const MeasureTableElement = (
     return measure.status ?? PageStatus.NOT_STARTED;
   };
 
+  const getCaption = (measureDisplay: string) =>
+    measureDisplay === "required"
+      ? "Required Measure Results"
+      : "Optional Measure Results";
```
benmartin-coforma: I think this would be fine as just a variable, since it will only change when the prop changes.
```suggestion
  const caption = table.measureDisplay === "required"
    ? "Required Measure Results"
    : "Optional Measure Results";
```

services/ui-src/src/components/export/ExportedReportWrapper.tsx
benmartin-coforma: I think we can simplify a fair bit here. You've got logic to ensure that we only assign a caption to the first element in each table, but I don't see a problem in assigning a caption to every element in the table. We will still only _use_ the first one, but it saves us a few conditionals in here.

There are also a couple superfluous checks & casts:
- `const sectionTitle = "title" in section ? (section.title as string) : undefined;`
   + As far as I can tell, no need to check whether title present, and no need to cast to string.
   + Simplifies to `const sectionTitle = section.title` with no effect (afaict)
   + And at that point, the variable is only saving us 1 character; we may as well not use it.
- `sectionTitle && section.type === "measure"`
   + I believe all of our Measure pages have titles?
   + And it's a good idea to use the PageType enum here
   + Simplifies to `section.type === PageType.Measure`
- `if (element.type === ElementType.SubHeader && "text" in element)`
   + I believe `text` is required for subheaders, so this can just check the type.
- `buildCaption(element.text as string)`
   + Since this is in a block where we know `element` is a subheader, we know the text is a string; no need to cast.

I also personally find it easier to reason about a variable like `mostRecentSubheader` vs `pendingCaption`. It's "the thing I just saw" vs "a thing I might use". So shifting perspectives there will let us centralize a little bit more logic in one spot: the function that determines the caption.

Put that all together, and we get something like this outside of the loop:
```tsx
  // Track SubHeader to use as caption for the pdf tables.
  let mostRecentSubheader: string | undefined = undefined;

  const determineCaption = (element: PageElement) => {
    if (!shouldUseTable(element)) {
      return undefined;
    } else if (mostRecentSubheader === undefined) {
      return section.title;
    } else if (section.type === PageType.Measure) {
      return `${section.title}: ${mostRecentSubheader}`
    } else {
      return mostRecentSubheader;
    }
  }
```

And something like this inside the loop.
```tsx
      const caption = determineCaption(element);

      if (element.type === ElementType.SubHeader) {
        mostRecentSubheader = element.text;
      }
```

I _think_ that has the same behavior overall?

services/ui-src/src/components/pages/Export/ExportedReportPage.test.tsx
benmartin-coforma: I don't think I have an opinion either way on what to assert - but if we _do_ use a loop, I much prefer `for (let element of elements) { ... }` to `elements.forEach(element => { ... })`

services/ui-src/src/components/export/ExportedReportElements.test.tsx
rocio-desantiago: True true

services/ui-src/src/components/report/WwlComponents/EligibilityTable.test.tsx
rocio-desantiago: I welcome the nitpicking! 😄

rocio-desantiago: Oh that's right only one will be visible 💡

services/ui-src/src/components/export/ExportedReportElements.test.tsx
rocio-desantiago: I tried `toHaveLength(2)` and that got funky but I was able to add `toBe(2)`

services/ui-src/src/components/export/ExportedReportTable.tsx
rocio-desantiago: For context: I did initially want to have the pattern that mimics MCR's use of captions with `TableContentShape` because we do have the `TableContentShape` interface but we're not using it. I tried using it and the `Page` elements started getting funky. It is doable, but it's worth discussing how our `PageElement` type is affected 🤗 I was going down the rabbit hole before I came to this alternative solution 😬 since the ticket was only calling for a few tables to get captions but of course I couldn't stop at the suggested tables and I kept going and here we are 👐 

In Table.tsx
```
interface Props {
  content: TableContentShape;
  variant?: string;
  children?: ReactNode;
}
```

in src/types/other.ts
```
export interface TableContentShape {
  caption?: string;
  headRow?: string[];
  bodyRows?: string[][];
  footRow?: string[][];
}
```

services/ui-src/src/components/pages/Export/ExportedReportPage.test.tsx
rocio-desantiago: No loop 😉

services/ui-src/src/components/export/ExportedReportWrapper.tsx
rocio-desantiago: Thanks! The only thing i had to update was `element` to be `element.type` in `if (!shouldUseTable(element)) {` and that works. I verified with the screenreader!

services/ui-src/src/components/report/MeasureTable.tsx
rocio-desantiago: Jes. I over-engineered that 😄 thanks haha

services/ui-src/src/components/export/ExportedReportElements.test.tsx
stacy-coforma: oh yeah duh that makes sense - checking `length` with `toHaveLength` is never going to work because we're looking at `length` as a number 🤪 yesterday was a long day haha

services/ui-src/src/components/export/ExportedReportElements.test.tsx
```diff
@@ -63,10 +63,10 @@ describe("Test ExportedReportElements", () => {
     const element = renderElements(section, mockNDR);
```
marcus-myers-co, line 66: This might be out of scope for this as it's a test file wide concern, but I noticed that we're not resetting the section elements between tests. Ideally no test should rely on another test or inflict potential side effects on other tests. My suggestion would be to add a beforeEach block that resets the section elements:
`beforeEach(() => {
  section.elements = [];
});`
Not a deal breaker, just a good practice to avoid state leakage between tests.

services/ui-src/src/components/report/WwlComponents/EligibilityTable.test.tsx
```diff
@@ -141,7 +141,8 @@ describe("<EligibilityTableElement />", () => {
   describe("EligibilityTableElementExport", () => {
     it("should render export table", () => {
       render(EligibilityTableElementExport(mockedElement));
-      expect(screen.getByText("mockTitle1")).toBeVisible();
+      expect(screen.getAllByText("mockTitle1").length).toBe(2);
```
marcus-myers-co: Small nitpick, but a suggested optimization:
```
const mockTitleElements = screen.getAllByText("mockTitle1");
expect(mockTitleElements).toHaveLength(2);
expect(mockTitleElements[0]).toBeVisible();
```
More of a personal pref, but if I'm going assert on an element and I don't expect it to change between assertions, I prefer todo only one DOM query. As for the assertion, .toHaveLength() is preferred as it will provide a better error message if it fails. .toHaveLength() is specifically for checking array / collection lengths and will show you the array it received if it fails which is helpful for debugging. Not worth holding up this PR, but just food for thought.

services/app-api/utils/reportValidation.ts
```diff
@@ -636,6 +637,7 @@ const reportValidateSchema = object().shape({
   year: number().required(),
   submissionCount: number().required(),
   archived: boolean().required(),
+  caption: string().notRequired(),
```
benmartin-coforma: Is this one a mistake? I don't remember seeing any report-level captions in this PR.

benmartin-coforma (dismissed):
All of my big concerns here have been addressed. Once you make decisions on my nitpick above, and on Marcus' comments, this PR has my approval to merge.

services/app-api/utils/reportValidation.ts
rocio-desantiago: Good catch. That looks like leftover code from a previous iteration. Removed!

services/ui-src/src/components/export/ExportedReportElements.test.tsx
rocio-desantiago: This is great feedback, thank you Marcus! Applied ✅

benmartin-coforma (dismissed):

cy-by (changes requested):
## QMS
✅  I re-reviewed QMS and table have `<captions>`.

## WWL
Since you added captions to the WWL report, I went through the flow and found that one table on the `/export` page had a `<caption>` but it is empty:

Image

benmartin-coforma (approved):

cy-by (approved):
Comments addressed! :shipit:

PR #630, opened 2026-03-25
CMDCT 5536 - New Waiver Demonstration Section Needs PDF Component

sopspops:
### Description

* Waiver demonstration section (both checkbox and list input) is added to PDF component
* Two sections added to PDF helper and component inventory updated

### Related ticket(s)

CMDCT-5536

---
### How to test

https://d1p8dscrfos2sn.cloudfront.net/
* Login as stateuser3
* Navigate to any form (for example - WWL)
* Check waivers and add waivers in list input
* Go to the review and check pdf and see waivers section is added

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


services/ui-src/src/components/export/ExportedReportWrapper.tsx
```diff
@@ -75,10 +76,25 @@ export const ExportedReportWrapper = ({ section }: Props) => {
...
+        const allCheckedChildren = answers.flatMap((answerValue) => {
+          const checkedChoice = element.choices.find(
+            (choice) => choice.value === answerValue
+          );
+          return checkedChoice?.checkedChildren ?? [];
+        });
+
+        // Keep checkbox element intact with original answers
+        return [element, ...expandCheckedChildren(allCheckedChildren)];
+      }
```
benmartin-coforma, line 91: A few things here.

---
The `Array.isArray()` check isn't needed here (and shouldn't ever be needed, except possibly in backend payload validation code). According to the [type definition for CheckboxTemplate](https://github.com/Enterprise-CMCS/macpro-mdct-hcbs/blob/8bc82f3ae219fef02cee754782d52ef8090e59c6/services/ui-src/src/types/report.ts#L401-L411), `element.answer` here will either be a `string[]` or `undefined`. You can verify this yourself by removing the function call: `const answers = element.answer ? element.answer : [];`

Then, if you hover over the `answers` variable, you'll see VSCode telling you that it's always a `string[]`. The only check we had to make was for truthiness.

string array type inference

---
And of course, you'll also see the complaint from our linter that the ternary operator is overkill here. All we need is the nullish coalesce operator: `const answers = element.answer ?? []`

But before we continue, consider that the `answers` variable is only used twice here: once for an early `return` check, and once to create the `allCheckedChildren` array. I do like the early return pattern for when a function gets complicated, but in this case I think removing it will actually simplify the function - because we'll be able to inline that variable without a big cost to readability:

```tsx
  if (element.type === ElementType.Checkbox) {
    const allCheckedChildren = (element.answer ?? []).flatMap(
      // ...
```

---
I would probably also inline the `checkedChoice` variable in that inner function. That will let us convert it from a `(foo) => { ...; return bar }` style function to just `(foo) => bar` style.

---
I also don't really understand what the comment about keeping the element intact means. At this point I'm getting really nitpicky, but I'd prefer to see a more explicit explanation of the render order.

---
Put that all together, and you get something like this:
```suggestion
      if (element.type === ElementType.Checkbox) {
        // Collect all checkedChildren from all selected answers
        const allCheckedChildren = (element.answer ?? []).flatMap(
          (answerValue) =>
            element.choices.find((choice) => choice.value === answerValue)!
              .checkedChildren ?? []
        );

        // The list of top-level answers is followed by the children of the 1st
        // selected answer, then the children of the 2nd selected answer, etc.
        return [element, ...expandCheckedChildren(allCheckedChildren)];
      }
```

benmartin-coforma: Oh, one more detail in there: I also changed the `find(...)?.checkedChildren` to `find(...)!.checkedChildren`. It shouldn't ever be possible for the list of answer IDs to contain an ID that doesn't match a choice value. If it does, I'd rather see the PDF view error out & us get a bug report, than for the page to quietly skip rendering an answer's children.

services/ui-src/src/components/fields/CheckboxField.tsx
```diff
@@ -119,6 +119,31 @@ export const CheckboxField = (props: PageElementProps<CheckboxTemplate>) => {
...
+
+  return (
+    <UnorderedList sx={sx.checkboxExport}>
+      {selectedLabels.map((label, i) => (
+        <ListItem key={`${label}-${i}`}>{label}</ListItem>
+      ))}
+    </UnorderedList>
+  );
+};
+
```
benmartin-coforma, line 136: There is also room for simplification here.

---
The `Array.isArray()` check in the initial `if` isn't doing anything. It can be removed with no effect on the function:
```ts
  if (!element.answer || element.answer.length === 0) {
    return <></>;
  }
```

---
The next `Array.isArray()` check is also redundant - especially because we've already ruled out `undefined`. Therefore, the `[String(element.answer)]` fallback will never be exercised and can also be removed:
```ts
  const selectedLabels = element.answer.map(
    (choiceId) =>
      element.choices.find((choice) => choice.value === choiceId)?.label ??
      choiceId
  );
```

---
Just like in ExportedReportWrapper, I would change the `find(...)?.label ?? id` to `find(...)!.label`. Because I think we're better off writing simple code that fails when our assumptions are wrong, than code which still kind of works when our assumptions are wrong. In this case, the code as-is "kind of works" in that that we would display an answer ID instead of a label, if no choice exists with that label.

But again, if our choices don't line up with our answer IDs, that's a bug I want to hear about as soon as possible, so that we can get to the root cause and fix it.

---
And now we're getting into really nitty nitpicks, but the function contains two `.map()` calls on the same sequence in a row, and we don't really use the intermediate result. That usually means it can be reduced down to a single `.map()`. We could extract the body of the first map as a standalone function, and call it from within the second.

Essentially, I'm saying `arr.map(x => foo(x)).map(y => bar(y))` can be rearranged to `arr.map(x => bar(foo(x)))`. Doing this doesn't always make things better, but in this case I think it helps.

Doing so also gives us a readily-available unique `key` for React, because answer IDs should always be unique.

---
Put that all together, and we get something like this:
```suggestion
export const CheckboxExport = (element: CheckboxTemplate) => {
  if (!element.answer || element.answer.length === 0) {
    return <></>;
  }

  const getLabel = (choiceId: string) =>
      element.choices.find((choice) => choice.value === choiceId)!.label

  return (
    <UnorderedList sx={sx.checkboxExport}>
      {element.answer.map((choiceId) => (
        <ListItem key={choiceId}>{getLabel(choiceId)}</ListItem>
      ))}
    </UnorderedList>
  );
};
```

services/ui-src/src/components/fields/ListInput.tsx
```diff
@@ -88,3 +94,39 @@ export const ListInput = (props: PageElementProps<ListInputTemplate>) => {
...
+  }
+
+  return (
+    <UnorderedList sx={sx.checkboxExport}>
+      {element.answer.map((item, i) => (
+        <ListItem key={`list-item-${i}`}>{item}</ListItem>
+      ))}
+    </UnorderedList>
+  );
+};
```
benmartin-coforma, line 110: Very nice 👍

sopspops:
@benmartin-coforma Thank you for the thorough review and explanation! It so helpful for me to understand the code better :) I applied all your feedback and it's ready for re-review!

benmartin-coforma (approved):
Looking good! Thanks Sophie!

stacy-coforma (approved):

PR #639, opened 2026-04-09
CMDCT 5732 - Critical Incidents - Page Titles

sopspops:
### Description

* Critical incidents page title name changes
* React helmet installed and created component with it
* tabTitle prop added
* Existing "title" prop name changed to "navTitle" since we have different title props, and this specific title is only for the navbar (Discussed with Ben!)

### Related ticket(s)

CMDCT-5732

---
### How to test

https://d1re4c5jqwgt2a.cloudfront.net/
* Navigate to Critical Incidents
* Create a form and see the title changes
* Enter a form and see page title changes on each section

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


services/ui-src/src/components/report/ReportPageWrapper.tsx
```diff
@@ -30,6 +30,8 @@ export const ReportPageWrapper = () => {
   } = useStore();
   const currentPage = useStore(currentPageSelector);
 
+  console.log("currentPage from report.pages:", currentPage);
```
rocio-desantiago: Ooh you probably don't need this anymore right? 🤔

services/ui-src/src/components/Title/index.tsx
```diff
@@ -0,0 +1,17 @@
+import { Helmet } from "react-helmet-async";
+
+interface TitleProps {
+  tabTitle: string;
+}
+
+export const Title = ({ tabTitle }: TitleProps) => {
+  const fullTitle = `${tabTitle}`;
```
rocio-desantiago: I don't remember why fullTitle was written as a separate variable. Would it still render if we update line 12 to 
> <title>{ `${pageTitle}` }</title>
If so, I'll need to make that change in QMR too.

rocio-desantiago (commented):
Two comments added

services/ui-src/src/components/Title/index.tsx
benmartin-coforma: Yes - in fact, I believe it would render the same without the string interpolation:
```tsx
<title>{tabTitle}</title>
```

sopspops: Thank you Rocio!! Just removed those and ready for re-review :)

rocio-desantiago: @benmartin-coforma thank you!

rocio-desantiago (dismissed):
LOOKS GOOD :D

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
```diff
@@ -121,174 +122,177 @@ export const DashboardPage = () => {
...
-                  everything looks good, confirm your entries and proceed.
-                </p>
-                <p>
-                  Once the report is generated, you can edit the name of the
-                  report and monitor its status in the dashboard below.
-                </p>
-                {reportType === ReportType.WWL ? (
+    <>
+      <Title tabTitle={`${reportName} - HCBS`} />
+      <PageTemplate type="report" sxOverride={sx.layout}>
```
benmartin-coforma: Request: Would you please swap these two lines? The `<Title>` will behave the same way when placed inside the `<PageTemplate>`. That will also let us remove the wrapping `<></>`, reducing the nesting level and the number of changed lines in this diff.

```suggestion
    <PageTemplate type="report" sxOverride={sx.layout}>
      <Title tabTitle={`${reportName} - HCBS`} />
```

benmartin-coforma (commented):
I've got good news and bad news.

Bad news first: we all told you that we needed to install a new library (`react-helmet`) in order to complete this story.

The good news is, we were probably wrong! It turns out that React 19 has built-in support for hoisting plain-old `<title>` tags up into the document head: https://react.dev/reference/react-dom/components/title

There are certain tags that `react-helmet` (or its friend, `react-helmet-async`) can handle that React cannot, but I don't think we need any of those.

And `react-helmet` supports tag deduplication: if we were to specify two `<title>`s, one in a parent component and one in a child component, the child's title would override the parent's. Whereas React would just throw _both_ titles into the document head, which is not kosher. But I think if we're careful we can avoid doing that.

Would you be willing to give this a try without the library? Everywhere we're using `<Title>`, you should be able to swap in a plain `<title>`, and everything should still work.

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
sopspops: Thank you Ben! I'll try without the library!!

sopspops: @benmartin-coforma So <title> is amazing! Got to remove Title from react-helmet and removed the package!

rocio-desantiago (dismissed):
Still working as expected ✅

cy-by (changes requested):
## 🚫 Multiple `<title>` tags

Starting at the CI dashboard page and moving forward into the CI form, there are multiple `<title>` tags, which [multiple title's go against the HTML spec](https://html.spec.whatwg.org/multipage/semantics.html#the-title-element:~:text=There%20must%20be%20no%20more%20than%20one%20title%20element%20per%20document).

The first title tag is the unique title for the page. The second title tag is the generic title tag from the home page.

Goal: there should only be one, unique `<title>...</title>` tag on the page.

HCBS CI Multiple title tags

services/ui-src/src/components/pages/Dashboard/DashboardPage.tsx
sopspops: @benmartin-coforma since using react-helmet seems like causing having two titles issues, I tried good old document.title, this way we can keep the react native title and prevent it having two titles, what do we think about this approach?

```
export const Title = ({ tabTitle }: TitleProps) => {
  const defaultTitle = useRef(document.title);

  useEffect(() => {
    document.title = tabTitle ?? defaultTitle.current;

    return () => {
      document.title = defaultTitle.current;
    };
  }, [tabTitle]);

  return null;
};
```

cy-by (approved):
Each page in the CI flow now has a unique title and there is only one `<title>` in the DOM :shipit:

rocio-desantiago (approved):
Checked the console this time and confirmed the title isn't showing up twice after clearning my cache 😄

benmartin-coforma (approved):
Works for me! Thanks for all your hard work on this, Sophie!

PR #658, opened 2026-04-28
CMDCT 5796 - QMS - Add Date Range to POM Measures

sopspops:
### Description

* Add date range to all POM measures
* DateField.tsx changed a bit to have two date fields, and end date error message added

### Related ticket(s)

CMDCT-5796

---
### How to test

https://d3cf7lx5w5kffp.cloudfront.net/
* Navigate to QMS and click any of POM measures
* Check there's a date range field
* Check an error message shows if end date is before the start date

Screenshot 2026-04-28 at 10 42 53 AM

### Notes

---
### Pre-review checklist

---
#### Review
- [x] Design: This work has been reviewed and approved by design, if necessary
- [x] Product: This work has been reviewed and approved by product owner, if necessary

#### Security


stacy-coforma:
I just wanted to assure you that I am looking at this 😅 since there is a lot going on and I'm less familiar with HCBS I'm taking my time a little and doing more investigation on how things work and looking at the patterns and everything.  I will have a review done for you hopefully today but more realistically tomorrow!

benmartin-coforma (commented):
Would you be willing to try an alternate implementation for this story?

I was imagining that both date fields would be combined in a single new component, DateRange.tsx, leaving DateField.tsx unchanged. The template (which should probably live in qmsElements.ts rather than elements.ts) might look like this:
```ts
export const measurementPeriodDateRange: DateRangeTemplate = {
  type: ElementType.DateRange,
  id: "measurement-period",
  labels: {
    top: "Measurement period dates",
    start: "Measurement start date",
    end: "Measurement end date",
  },
  helperText:
    "Select the measurement period start and end dates for this individual metric.",
  required: true,
};
```

And the answer might look like `answer?: { start? string; end?: string; };`

This would make the inter-field validation (start <= end) less awkward; there wouldn't be a special need to pass an `allElements` prop. And it would be the same number of new element types, since there would be no need for the new `Label` element.

It _would_ be more new code overall though; the DateRange component would need to manage validating & displaying errors for multiple fields. You might find the NDR component helpful as an example of how we've done that elsewhere. It would also require new logic in `completeness.ts`.

And either way, we will want new documentation in `docs/ReportSchema.md`.

I _think_ this approach will feel nicer overall? But I'm not totally sure. Worth a shot?

sopspops:
@benmartin-coforma Thank you! This totally makes sense and looks cleaner. I was torn between editing dateField or creating another component, I went with adding less code, but this makes more sense! I reversed dateField and added dateRange, how does this look?

services/ui-src/src/constants.ts
```diff
@@ -1,5 +1,5 @@
 export const ErrorMessages = {
-  endDateBeforeStartDate: "End date can't be before start date",
+  endDateBeforeStartDate: "Measurement end date can't be before start date",
```
benmartin-coforma: Where else in the app is this constant used? Is this change appropriate there?

services/ui-src/src/components/report/Page.test.tsx
```diff
@@ -276,6 +276,30 @@ describe("Page Component with state user", () => {
...
+        ]}
+        setElements={setElements}
+      />
+    );
+
+    const dateField = screen.getByRole("textbox");
+    await userEvent.type(dateField, "10162024");
+
+    expect(setElements).toHaveBeenCalled();
+  });
```
benmartin-coforma, line 311: This is a nice test. My first impulse was to say that we're already testing this in DateField.test.tsx, but that's only partially true. The thing that this test verifies is that the callback propagates back up through Page.
1. Can you rename the test to reflect that? Perhaps `"should transmit changes to its parent through setElements"`?
2. Can you make the assertion more specific? What should `setElements` have been called with?

services/app-api/types/reports.ts
```diff
@@ -385,11 +387,27 @@ export type DateTemplate = {
   type: ElementType.Date;
   id: string;
   label: string;
-  helperText: string;
+  helperText?: string;
```
benmartin-coforma: If all elements of this type have `helperText` (and I think they do), then it should be required. Type definitions are a way of communicating info throughout the app, and "you will always have helperText" is info worth communicating 👍 

The same goes for the DateRangeTemplate type definition. Will also mean changing the type definitions in ui-src, and changing the reportValidation code.

services/ui-src/src/components/fields/DateRange.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const rangeErrorMessage = (() => {
+    const parsedStart = parseMMDDYYYY(dateRange.answer?.start ?? "");
+    const parsedEnd = parseMMDDYYYY(dateRange.answer?.end ?? "");
+
+    if (parsedStart && parsedEnd && parsedEnd < parsedStart) {
+      return ErrorMessages.endDateBeforeStartDate;
+    }
+
+    return "";
+  })();
```
benmartin-coforma: I think having an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) in a React component is a bit unusual. It would feel a little bit more natural, IMO, to `let rangeErrorMessage = ""` and modify it if necessary.

But even then, we would have redundant parsing code: each date is parsed by its respective change handler, and then parsed again here on every render. The way we solve that problem elsewhere in the app is by combining change handlers. One example is in NDR.tsx, which parses numbers & possibly displays errors based on their values. Another is AdminBannerForm.tsx, which actually has the same "end date before start date" validation needed here.

Both of those examples also combine display values and error messages into single objects; see the comment above `makeEmptyStringCopyOf()` for an explanation of how. I don't know if that approach is really needed here, since there are only two fields (start date and end date). But switching to that approach _would_ make this component line up more closely with our existing code.

Can you take a look and see what you think?

benmartin-coforma (commented):
This is looking pretty good! Thanks for trying out the combined component approach.

Question: If you fill out only the start date (but not the end date), can the measure still be marked as complete? I suspect the answer will be "yes", because I think completeness.ts needs an update.

services/ui-src/src/constants.ts
sopspops: It is only used for dateRange! I'll move the message to dateRange!

services/ui-src/src/components/fields/DateRange.tsx
sopspops: That makes sense. Thank you Ben!! I'll match this logic with other ones!

sopspops: @benmartin-coforma All the feedback applied! :)

benmartin-coforma (commented):
This code looks fine, but if I:
1. Fill out a POM measure, selecting FFS
2. Fill out the FFS sub page
3. Delete the start date and/or end date
I am still able to mark the measure as completed. I shouldn't be able to.

This is because completeness.ts is only checking to see if the answer field is present (and not checking to see if it has nonempty `.start` and `.end` properties). This component will need a custom check there - just like the custom check for NDR.

services/ui-src/src/components/fields/DateRange.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+    const nextAnswer = {
+      ...dateRange.answer,
+      [fieldName]: isValid ? maskedValue : undefined,
+    };
+    const rangeError = getRangeErrorMessage(nextAnswer);
+
+    updateElement({
+      answer: nextAnswer,
+    });
```
benmartin-coforma:
```suggestion
    if (!rangeError) {
      updateElement({ answer: nextAnswer });
    }
```

services/ui-src/src/utils/state/reportLogic/completeness.ts
```diff
@@ -250,6 +251,18 @@ export const elementSatisfiesRequired = (
...
+  if (element.type === ElementType.DateRange) {
+    const start = element.answer?.start?.trim();
+    const end = element.answer?.end?.trim();
+    if (!start || !end) return false;
+
+    const parsedStart = parseMMDDYYYY(start);
+    const parsedEnd = parseMMDDYYYY(end);
+    if (!parsedStart || !parsedEnd) return false;
+
+    return parsedEnd >= parsedStart;
```
benmartin-coforma:
```suggestion
    return !!element.answer.start && !!element.answer.end;
```

services/ui-src/src/utils/state/reportLogic/completeness.test.ts
```diff
@@ -486,6 +487,47 @@ describe("elementSatisfiesRequired", () => {
...
+  });
+
+  test.each([
+    undefined,
+    {},
+    { start: "2026-01-01" },
+    { end: "2026-01-31" },
+    { start: "", end: "2026-01-31" },
+    { start: "2026-01-01", end: "" },
+    { start: "05/12/2026", end: "05/11/2026" },
```
benmartin-coforma:
```suggestion
```

benmartin-coforma (commented):
I apologize for the delayed response on this, and for this yet-another change request.

I have 3 code suggestions below, which add up to this: Let's never store invalid data. If the DateRange component has strict validation up front, then no other code has to check whether the data is _valid_ - only whether it is _present_. This both simplifies code and gives us more confidence in the data.

services/ui-src/src/utils/state/reportLogic/completeness.test.ts
sopspops: @benmartin-coforma Thank you again for reviewing this PR😭 and thank you for the detailed explanations- it is so helpful to me!! Applied all these changes!!

benmartin-coforma (approved):
Approved! Thank you so much for all the work on this!

stacy-coforma (approved):
Thanks @benmartin-coforma for reviewing so thoroughly!  I gave everything a look and tested in the deployed environment, looks like it is working well and rendering on the PDF export as expected 👍

PR #662, opened 2026-04-29
[HCBS] Notifications - Send email on status change - PROOF OF CONCEPT

rocio-desantiago:
### Description
This PR includes a detailed email being sent to the report user based on the email address inserted in the `General Information` page. Additionally, we have a button to test sending an email by just entering your personal email directly via the Admin Notifications page. This is covered behind the `notificationsSystem` flag. The report status should also only work with `isDev` and not in val and prod environments. If/when we are ready to push to val and prod, we will have to resubmit DNS domain names for those environments the way we requested for the dev environment.

For reference: [Documentation in Confluence](https://confluenceent.cms.gov/spaces/CMDCT/pages/1365428814/Amazon+SES+in+MDCT+Apps)
Related PR over on RHTP: [Email notifications]( https://github.com/Enterprise-CMCS/macpro-mdct-rhtp/pull/184)

Email sent
- In Progress -> Submitted (a user submits a report)
- Submitted -> In Progress (an admin user unlocks a report)

Example Email from Status Change
Example Email sent to user upon report status change

Admin Notifications Page with button to send Test Email 
Admin Notifications Page with button to send Test Email
Example Sending Test Email from Admin Notifications Page
Test Email example

### Related ticket(s)
CMDCT-5573

---
### How to test
Deploy Link: https://d1o2sijmqgzsmo.cloudfront.net/

**As a state user**
1. Create a report (any)
2. On the general information page, enter your email
3. Fill out and submit
4. Verify you receive "Submitted" Status Email

**As an admin user (adminuser@test.com):**
1. Search for user report in report dashboard
2. Unlock the report
3. Verify you receive "In Progress" status email

**Bonus for admin user (adminuser@test.com):**
1. Click on Admin/Notifications
2. Enter your email in "Enter recipient email"
3. Click "Send Test Email"
4. Verify you receive automated test email :tada:

### Notes

> In LocalStack, SES EmailIdentity is skipped (it fails because LocalStack can't verify cms.hhs.gov) and the SES policy uses "*" instead. In real AWS deployments the domain identity is created and the policy is scoped to it.

- [x] SES in production mode via Kion
  > aws sesv2 put-account-details \
  > --production-access-enabled \
  > --mail-type TRANSACTIONAL \
  > --website-url https://mdcthcbsdev.cms.gov/ \
  > --additional-contact-email-addresses rocio.de-santiago@coforma.io \
  > --use-case-description "To send notification emails to CMS users for HCBS reports that switch from status: In Progress to Submitted status" \
  > --contact-language EN
- [x] Settings in code pushed to local branch: poc-ses
    ```
    new ses.EmailIdentity(scope, "SenderDomainIdentity", {
      identity: ses.Identity.domain("cms.hhs.gov"),
    });
    ```
- [x] DKIM setup and csv with CNAME values generated
    Identity verification pending
- [x] HCBS_Internet_DNS_Change_Request.pdf: Signed by Junni and CMS Group Director
- [x] [CMS Service Desk request for an Internet DNS Change](https://jira.cms.gov/servicedesk/customer/portal/1/create/291) Help desk ticket submitted with Signed PDF and csv file
- [x] [WAKA-21376 - Add CNAME records to cms.hhs.gov](https://jira.cms.gov/browse/WAKA-21376) ticket for the Akami team to handle this request

---
### Pre-review checklist

---
#### Review
- [x] Design: This work has been reviewed and approved by design, if necessary
- [x] Product: This work has been reviewed and approved by product owner, if necessary


cli/commands/local.ts
```diff
@@ -96,6 +97,13 @@ export const local = {
 
     await seedData();
 
+    const sesClient = new SESClient({ region });
+    await sesClient.send(
+      new VerifyEmailIdentityCommand({
+        EmailAddress: "MDCT_NoReply@cms.hhs.gov",
```
stacy-coforma: Since we don't have access to this inbox, I wonder what is getting sent to it 🤔 may need to test with a different sender locally/in dev that we can control for the purposes of testing the proof of concept.  Not sure if this makes sense to make an env variable since in real life we'll probably consistently want to use this address, though. Ultimately we will want to use this address, so we will have to go through the process of getting it verified with Junni or someone else who can access that inbox at some point.

deployment/stacks/api.ts
```diff
@@ -140,6 +140,12 @@ export function createApiComponents(props: CreateApiComponentsProps) {
     }),
   ];
 
+  const sesPolicy = new iam.PolicyStatement({
+    effect: iam.Effect.ALLOW,
+    actions: ["ses:SendEmail", "ses:SendRawEmail"],
+    resources: ["*"],
```
stacy-coforma: Fine for local testing; we should be more restrictive with the IAM policy when it gets to real environments.

services/ui-src/src/components/pages/Admin/NotificationsPage.tsx
```diff
@@ -8,12 +8,29 @@ import {
...
   const [notifications, setNotifications] = useState<Notification[]>([]);
   const [loading, setLoading] = useState(true);
+  const [sending, setSending] = useState(false);
+  const { notificationsSystem } = useFlags() ?? {};
+
+  const handleSendEmail = async () => {
+    setSending(true);
+    try {
+      await sendEmail({
+        toAddress: "test@test.com",
```
stacy-coforma: would like to be able to test this to an inbox we have access to to ensure it's working 👀 I'm not sure if the dev instances that get spun up per PR will have actual functionality potentially; if so we might want to consider having an input field to send a test email instead of hardcoding one so folks can test on inboxes they can see

services/ui-src/src/components/pages/Admin/NotificationsPage.tsx
```diff
@@ -74,13 +91,28 @@ export const NotificationsPage = () => {
               }
             />
           ))}
+          <Box mt="spacer4">
+            {notificationsSystem && (
+              <Button
```
stacy-coforma: I am not able to actually see the button 😞 CORS errors on my side.  I know you mentioned you'd solved them locally?  Wondering why that hasn't translated to my local instance 🫠 we can sync up on that when you are back from CfA

stacy-coforma (commented):
After some trial and error locally, I was able to get localstack to confirm it is "sending" the message (simulating at least)

Image

I did have trouble with the test button locally and left a note on that.

To call this concept "proofed", I think we will have to see if any real functionality is possible in the ephemeral environment for this PR (probably will be easiest with a sender inbox we can control for POC purposes).  Once we confirm that, we can move on to getting Junni's help (or anyone else who has access to that inbox; currently I don't know of anyone else who does) with getting the actual NoReply sender address validated for use with the app.

I left notes all over the place as mostly thought dumps while I was looking through 😅 happy to talk about anything I wrote in here and assist with the remainder of the journey.

services/ui-src/src/components/pages/Admin/NotificationsPage.tsx
rocio-desantiago: oooh I like the idea for an input field so we can test different inboxes receiving the email

cli/commands/local.ts
rocio-desantiago: steps verified for this and added to notes section ✅

deployment/stacks/api.ts
rocio-desantiago: Would it be the `effect:` that gets updated to be more restrictive?

stacy-coforma: I *think* (definitely am going to need to research to confirm) that the `effect` is okay, but `resources: ["*"]` is too permissive.  I'm sure there are some specific things we should be putting in there (but again am going to have to research to confirm what those things are; sorry for leaving a vague response I just felt bad that you asked this before I was OOO and I didn't reply 😅)

services/ui-src/src/components/pages/Admin/NotificationsPage.tsx
rocio-desantiago: Confirming that this will only work in deploy and not localhost, now that the domain has been verified with dev via AWS and CMS IT.

services/ui-src/src/components/pages/Admin/NotificationsPage.tsx
```diff
@@ -74,13 +88,41 @@ export const NotificationsPage = () => {
               }
             />
           ))}
+          <Box mt="spacer4">
+            {notificationsSystem && (
```
ntsummers1: Super, super small nit: If we've got the feature flag here I think we can swap these two so that <Box mt="spacer4"> only renders if that featureflag is turned on

services/app-api/handlers/reports/submit.ts
```diff
@@ -57,5 +58,12 @@ export const submitReport = handler(parseReportParameters, async (request) => {
   // save the report that's being submitted (with the new information on top of it)
   await putReport(validatedPayload);
 
+  try {
+    await sendEmail(report);
+  } catch (error) {
+    // log and allow call to succeed even if email fails
+    logger.error(error);
+  }
+
```
ntsummers1: This is ok for now. Not a blocker. But some sort of alert to let us or the user know an email wasn't sent could be helpful.

ntsummers1 (dismissed):
This is super awesome! I was able to confirm I got all 3 emails, 1 for the state and 2 for both admin paths. I left 2 comments that aren't blocking at all. It looked like the feature flags and isDev checks should ensure this only shows in dev too.

services/app-api/handlers/reports/submit.ts
rocio-desantiago: To-do added 🎉

sopspops (dismissed):
Tested and it works and looks good! :)

services/ui-src/src/components/pages/Admin/NotificationsPage.tsx
rocio-desantiago: Good call! Updated ✅

stacy-coforma (dismissed):
Putting a request changes on just to prevent accidental merges 😅 

While testing we noticed that the feature flag is not correctly toggling the behavior on submit/reopen, so we're looking at either adding a flag check in the lambda or moving the call for sending the email to the frontend where the flag check is already set up.

deployment/stacks/api.ts
```diff
@@ -110,6 +113,7 @@ export function createApiComponents(props: CreateApiComponentsProps) {
       tables.map((table) => [`${table.node.id}Table`, table.table.tableName])
     ),
     brokerString,
+    ...(LD_SDK_SERVER ? { LD_SDK_SERVER } : {}),
```
benmartin-coforma: It should be fine to pass this straight through, right?
```suggestion
    LD_SDK_SERVER,
```

deployment/stacks/api.ts
```diff
@@ -213,13 +242,16 @@ export function createApiComponents(props: CreateApiComponentsProps) {
...
+  if (isDev) {
+    new Lambda(scope, "submitReport", {
+      entry: "services/app-api/handlers/reports/submit.ts",
+      handler: "submitReport",
+      path: "reports/submit/{reportType}/{state}/{id}",
+      method: "POST",
+      ...commonProps,
+      additionalPolicies: [...additionalPolicies, sesPolicy],
+    });
+  }
```
benmartin-coforma: We still want to create the lambda in non-dev environments, right? Same for releaseReport, below.
```suggestion
  new Lambda(scope, "submitReport", {
    entry: "services/app-api/handlers/reports/submit.ts",
    handler: "submitReport",
    path: "reports/submit/{reportType}/{state}/{id}",
    method: "POST",
    ...commonProps,
    additionalPolicies: isDev ? additionalPolicies : [...additionalPolicies, sesPolicy],
  });
```

services/app-api/libs/ses-lib.ts
```diff
@@ -0,0 +1,20 @@
...
+  endpoint: process.env.AWS_ENDPOINT_URL,
+  forcePathStyle: true,
+};
+
+const client = new SESClient(awsConfig);
+
+export default {
+  sendSesEmail: (params: SendEmailRequest) =>
+    client.send(new SendEmailCommand(params)),
+};
```
benmartin-coforma: Nitpick: I'd probably use the simpler style export here. I don't think defining a default export gains us much, and the syntax is a little simpler without it.

```suggestion
export const sendEmail = (params: SendEmailRequest) =>
    client.send(new SendEmailCommand(params));
```

services/app-api/libs/localstack.ts
```diff
@@ -0,0 +1,12 @@
+export const isLocalStack = (): boolean => {
+  return (process.env.AWS_ENDPOINT_URL &&
+    process.env.AWS_ENDPOINT_URL.includes("localhost")) as boolean;
+};
```
benmartin-coforma: Nitpick: A safe access operator + nullish coalesce can save a tiny bit of code here.

```suggestion
export const isLocalStack = () =>
  process.env.AWS_ENDPOINT_URL?.includes("localhost") ?? false;
```

As an aside, it's usually not necessary to annotate a function's return type:
```ts
const foo = (): SomeType => { ... };
//              ^^^^^^^^
```
Typescript should know whether the function body returns `SomeType` or not, without you explicitly saying so. The only time I add this annotation is when I'm writing a function that is both 1) complicated, and 2) likely to be modified, in a way that might change the return type. That way, when some future dev changes the function, TS will tell them right away that if they need to button things up. But I think in most cases we're better off just letting the types flow.

As another aside, you're already aware it's not usually necessary to cast, especially the type of a value being generated from your own code. For example:
```ts
type FooType = { bar?: string; }
const isQuux = (foo: FooType) => foo.bar && foo.bar.includes("quux");
const isFred = (foo: FooType) => !!foo.bar && foo.bar.includes("fred");
```
The return type of `isQuux` is `boolean | undefined`, because when `bar` is undefined, the `&&` short-circuits, the boolean `.includes()` is never evaluated, and the `undefined` flows through.

Whereas the return type of `isFred` is just `boolean`, because when `bar` is undefined, value of `!!foo.bar` is `false` the same short-circuit happens, and the `false` flows through. So, if you prefer this change to the `?.` and `??` above, it gets you to the same place.

services/app-api/libs/localstack.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const fixLocalstackUrl = (url: string): string => {
+  if (isLocalStack()) {
+    return url.replace(".localstack.cloud", "");
+  }
+
+  return url;
+};
```
benmartin-coforma: Surely we will only have `".localstack.cloud"` in a URL if it is a localstack URL? If so, this simplifies to
```suggestion
export const fixLocalstackUrl = (url: string) =>
  url.replace(".localstack.cloud", "");
```

services/app-api/utils/notifications/email.ts
```diff
@@ -0,0 +1,77 @@
...
+  );
+  if (
+    !contactEmailElement ||
+    !("answer" in contactEmailElement) ||
+    !contactEmailElement.answer
+  ) {
+    return [];
+  }
+  return [contactEmailElement.answer as string];
+};
```
benmartin-coforma: Suggestion: This gets the job done, but I think you're suffering from the interface you chose. If instead of returning a possibly-empty string array, this function were to return a possibly-falsy string, then the early returns could be replaced by safe-access operators.

In fact! You _could_ do it without intermediate variables, in a single chain of `.find()` and `.filter()` calls, although you'll have to discover a little trick in order to also avoid casts.

The trick is that Typescript _can_ be clever about narrowing types during a `.filter()`, but only in the most obvious cases. [Here's an illustration in the Typescript Playground](https://www.typescriptlang.org/play/?strictBuiltinIteratorReturn=true#code/C4TwDgpgBAKgTgSwIYDsDmAbaBeKBvKAZwQBMJCAuKAZgBook4IkqUBXAWwCMI4oBfANwAoUJCgAlCAGNgqTDnxFS5KgBZ6jZq048+Q0eGgBlABZJxueMnRYoAH0ky5tiCOEAzNilkIA9ihQAGIIKCQAQgho1vJYhAD6AIxQABSE5pCUUGYWEADaALoAlPjCUFBMwGxwgem5hAB0HggYwLxpplDYAHxEpg3EZIRd2LjUUABkE30NWkhQvQCsRSL8wp7evgHBoRFRMa4JAEypdZlUOZCFJXhlFRBVNX31TS1tcB1dvekDKsOjYyKr1a7XSXxmcwWUGWq3W0gChGAz3O2Qy+QKXSgeTuBEGqhomiYLCgAHYBLQccohupCdooAAOcmUvFZOgMIlUE5rAoiLBIjggaxZA4KQruAD04tgRkI0kQYCRAFE4HA-HAqAByKSyWIQDVQBDDFB+JFIQjENAoJBcOzAPxQMTQDUirAa4QCoWYkJhSLRRC6hKJDr1FbrSVQADqaoA1sMAFZsRFQZooCDuwWIf47H37f2HeJHYOZUNAA). If you can see why Typescript is able to make the necessary inferences for one of those functions and not for the other, then you can solve the puzzle to make this function into a single chain.

Whether this function would be _better_ as a single chain is another question entirely; maybe worth trying out but not a given. But I do think it would be better returning `string | undefined`, unless you have a compelling reason otherwise.

services/app-api/utils/notifications/email.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export const sendEmail = async (report: Report) => {
+  const { name, status } = report;
+  const recipients = getRecipients(report);
+  if (recipients.length === 0) return;
```
benmartin-coforma: Since we're still in the Proof of Concept phase, this is probably fine, but sooner or later we should probably at least log a warning when trying to send an email for a report with no contact.

services/app-api/libs/launchdarkly-lib.ts
benmartin-coforma, line 1: I did talk about potentially using the LaunchDarkly API directly, and I do think that is feasible especially on the backend here, but I don't want to spend any of your time trying to make that happen. Maybe a fun little project for later :)

services/app-api/utils/notifications/email.test.ts
```diff
@@ -0,0 +1,29 @@
+import { sendEmail } from "./email";
+import sesLib from "../../libs/ses-lib";
+import { validReport } from "../tests/mockReport";
+
+jest.mock("../../libs/ses-lib", () => ({
+  __esModule: true,
+  default: {
+    sendSesEmail: jest.fn(),
+  },
+}));
```
benmartin-coforma: Using a named export in ses-lib will make this mock a little simpler 👍

services/app-api/utils/notifications/email.test.ts
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  test("should not issue a send email command when no email in report", async () => {
+    await sendEmail(validReport);
+    expect(sesLib.sendSesEmail).not.toHaveBeenCalled();
+  });
+
+  test("should issue a send email command when email in report", async () => {
+    // any type so it doesn't complain about accessing .answer on generic PageElement
+    const reportWithEmail: any = structuredClone(validReport);
```
benmartin-coforma, line 19: A very valid use of `any` in my opinion.

services/app-api/handlers/reports/release.test.ts
```diff
@@ -95,4 +104,14 @@ describe("Test releaseReport handler", () => {
 
     expect(res.statusCode).toBe(StatusCodes.Ok);
   });
+
+  test("Test release succeeds without sending email when flag is off", async () => {
+    (getFlag as jest.Mock).mockResolvedValueOnce(false);
+    const { sendEmail } = require("../../utils/notifications/email");
```
benmartin-coforma: Suggestion: Would prefer to see this `require` as an `import` at the top of the file instead.

services/ui-src/src/utils/api/requestMethods/notifications.test.ts
```diff
@@ -30,4 +34,16 @@ describe("utils/notifications", () => {
...
+    test("executes", () => {
+      expect(
+        sendTestEmail({
+          toAddress: "test@example.com",
+          subject: "Test Subject",
+          message: "Test message",
+        })
+      ).toBeTruthy();
+    });
+  });
```
benmartin-coforma, line 57: I'm a little confused on this test - although I see it does follow the pattern of other tests in the file.
* I don't see where / how we're mocking the API. Does this actually try to send network traffic during `yarn test`?
* `sendTestEmail()` is async, therefore returns a `Promise`, and promises are objects which are always truthy. Should there be an assertion that we called the right endpoint with the right arguments?

benmartin-coforma (commented):
I've got so many comments in here I don't even remember whether they're all nitpicks or if I found something blocking. Lo siento T_T

deployment/stacks/api.ts
rocio-desantiago: non-dev environments as in val and prod? Not yet.

services/app-api/utils/notifications/email.test.ts
rocio-desantiago: true true

services/app-api/handlers/reports/release.test.ts
rocio-desantiago: true true

services/app-api/libs/localstack.ts
rocio-desantiago: You right 😄

deployment/stacks/api.ts
rocio-desantiago: testing 1, 2, 3

services/app-api/libs/localstack.ts
rocio-desantiago: 😬

services/ui-src/src/utils/api/requestMethods/notifications.test.ts
rocio-desantiago: Gonna have to work on this a little bit. 

1. Is network traffic being made?
No. setupJest.tsx globally mocks aws-amplify/api for the entire ui-src test suite, so no real requests are made.
2. Good point. Gonna see if i can assert that apiLib.post was called with the right path and body instead.

services/app-api/libs/localstack.ts
rocio-desantiago: updated to 
```
export const fixLocalstackUrl = (url: string) =>
  isLocalStack() ? url.replace(".localstack.cloud", "") : url;
```
because it was failing as such:

>     expect(received).toBe(expected) // Object.is equality
> 
>     Expected: "https://bucket.s3.localstack.cloud/key"
>     Received: "https://bucket.s3/key"
> 
>       17 |     process.env.AWS_ENDPOINT_URL = undefined;
>       18 |     const url = "https://bucket.s3.localstack.cloud/key";

deployment/stacks/api.ts
benmartin-coforma: I understand we don't want emails in val & prod yet, but this `if` will prevent the submit endpoint from being created altogether. I don't think that's what we want.

services/app-api/libs/localstack.ts
benmartin-coforma: I don't mind this, but I do think that's a bad test. "If a certain environment variable is not set, then this function should not do what it's named" doesn't seem like behavior we want, and therefore it's not behavior we should be asserting on.

services/app-api/libs/launchdarkly-lib.ts
```diff
@@ -0,0 +1,27 @@
+import * as ld from "@launchdarkly/node-server-sdk";
+
+let client: ld.LDClient | undefined;
+
+async function getLdClient(): Promise<ld.LDClient> {
+  if (client) return client;
+  const sdkKey = process.env.LD_SDK_SERVER ?? "local";
+  client = ld.init(sdkKey);
```
stacy-coforma: We need to also set the URLs to be the govcloud endpoints so that our SDK key can authenticate:

```suggestion
  client = ld.init(sdkKey, {
    baseUri: "https://sdk.launchdarkly.us/",
    eventsUri: "https://events.launchdarkly.us/",
    streamUri: "https://stream.launchdarkly.us/",
  });
```

deployment/stacks/api.ts
rocio-desantiago: oh no that's not what we want 😭

deployment/stacks/api.ts
rocio-desantiago: Correct ✅

services/app-api/utils/notifications/email.ts
rocio-desantiago: Ah just now seeing this. Will noodle on this

deployment/stacks/api.ts
rocio-desantiago: I lied. That failed. 🫠 
CDK's Lambda construct throws if an environment variable value is undefined. In the LocalStack case, LD_SDK_SERVER is not in the props and will be undefined, which causes a synth error. 

_For my own knowledge: CDK synth is the step where CDK coverts our TypeScript code into a CloudFormation template._

services/app-api/libs/launchdarkly-lib.ts
rocio-desantiago: Applied ✅ Thank youuuu

deployment/stacks/api.ts
rocio-desantiago: because of the error, I updated to this 
```
    ...(launchDarklyServer && { LD_SDK_SERVER: launchDarklyServer }),
```

deployment/stacks/api.ts
rocio-desantiago: isDev is not even needed 😄 the ld flag takes care of it 🙌

services/app-api/libs/localstack.ts
rocio-desantiago: Thank you 🙏

services/app-api/libs/localstack.ts
rocio-desantiago: That's a good point. localstack is still so confusing to me 🤭

services/ui-src/src/utils/api/requestMethods/notifications.test.ts
rocio-desantiago: Updated 🙌

services/app-api/utils/notifications/email.ts
rocio-desantiago: That's insane that TypeScript can do that 🤯 
I have a theory to test, but will wait for flag to be back online to test 😉

rocio-desantiago: It works! Please review at your leisure @benmartin-coforma

deployment/stacks/api.ts
benmartin-coforma: Thanks for trying that out!

deployment/stacks/api.ts
```diff
@@ -110,6 +113,7 @@ export function createApiComponents(props: CreateApiComponentsProps) {
       tables.map((table) => [`${table.node.id}Table`, table.table.tableName])
     ),
     brokerString,
+    ...(launchDarklyServer && { LD_SDK_SERVER: launchDarklyServer }),
```
stacy-coforma: I feel tempted to suggest not re-mapping this to `LD_SDK_SERVER` because within secrets manager it is already defined as `launchDarklyServer` 🤔 I'll go through the rest of it and add suggestions on other spots that would need to be modified to support this change.  I know this is not the conventional way that local env variables look, but this is how they are set within AWS so I feel it might make sense to stick with that for consistency.

```suggestion
    launchDarklyServer,
```

services/app-api/libs/launchdarkly-lib.ts
```diff
@@ -0,0 +1,31 @@
+import * as ld from "@launchdarkly/node-server-sdk";
+
+let client: ld.LDClient | undefined;
+
+async function getLdClient(): Promise<ld.LDClient> {
+  if (client) return client;
+  const sdkKey = process.env.LD_SDK_SERVER ?? "local";
```
stacy-coforma: I think this is the only other spot that would need updating if we change the variable passed at the lambda 👀 

```suggestion
  const sdkKey = process.env.launchDarklyServer ?? "local";
```

services/app-api/libs/tests/launchdarkly-lib.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    expect(result).toBe(false);
+  });
+
+  it("falls back to 'local' when LD_SDK_SERVER is not set", async () => {
+    mockVariation.mockResolvedValue(false);
+
+    await getFlag("testFlag");
+
+    expect(mockInit.mock.calls[0][0]).toBe("local");
```
stacy-coforma: This works as-is, but I think something like this might be more readable:

```suggestion
    expect(mockInit).toHaveBeenCalledWith("local", expect.anything());
```

or we could use `toHaveBeenNthCalledWith` to specifically look for the first (only?) call.

services/app-api/libs/tests/launchdarkly-lib.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  it("returns defaultValue when LaunchDarkly client fails to initialize", async () => {
+    mockWaitForInitialization.mockRejectedValueOnce(new Error("timeout"));
+
+    const result = await getFlag("testFlag", false);
+
+    expect(result).toBe(false);
+  });
+
+  it("falls back to 'local' when LD_SDK_SERVER is not set", async () => {
```
stacy-coforma, line 59: I think this is actually the only case being tested here; we are not setting `LD_SDK_SERVER` in this test so I believe all of these are going to be defaulting to `local` for the key.  But if we were to test that this is another spot that would need to be updated if we changed it to `launchDarklyServer`.

```suggestion
  it("falls back to 'local' when launchDarklyServer is not set", async () => {
```

services/app-api/libs/localstack.ts
```diff
@@ -0,0 +1,5 @@
+export const isLocalStack = () =>
+  process.env.AWS_ENDPOINT_URL?.includes("localhost") ?? false;
+
+export const fixLocalstackUrl = (url: string) =>
```
stacy-coforma: Is this still used anywhere?  I only see it in the `localstack.test.ts` file but may be missing something?

services/app-api/utils/notifications/email.ts
```diff
@@ -0,0 +1,71 @@
...
+    });
+    return;
+  }
+  const emailTemplate = getTemplate(name, status, [recipient]);
+  logger.info("Sending email to: ", recipient, "with content: ", emailTemplate);
+  if (!isLocalStack()) {
+    await sesLib(emailTemplate);
+  } else {
+    logger.info("Skipping email in dev env");
+  }
```
stacy-coforma: I'm not sure if this is necessary; Localstack is able to mock SES functionality, so I don't think we necessarily need to filter to is/is not localstack.  In fact it would let us look at the message it would have sent (though it won't actually trigger an email) so it could be good to let it simulate in a local environment.

deployment/local/prerequisites.ts
```diff
@@ -35,6 +35,9 @@ export class LocalPrerequisiteStack extends Stack {
         vpcName: SecretValue.unsafePlainText("localstack-dev"),
         brokerString: SecretValue.unsafePlainText("localstack"),
         kafkaAuthorizedSubnetIds: SecretValue.unsafePlainText(subnet1.subnetId),
+        launchDarklyServer: SecretValue.unsafePlainText(
+          process.env.LD_SDK_SERVER ?? ""
```
stacy-coforma, line 39: We'll also need to update `.env.tpl` with this value so it can pull it out of 1password

```env
LD_SDK_SERVER=op://mdct_devs/hcbs_secrets/LD_SDK_SERVER
```

(Sorry if this was done elsewhere and I missed it!)

stacy-coforma (changes requested):
I left some comments!  Much of this is optional, but a couple things I think we should do before merging:

1. Add the `LD_SDK_SERVER` env variable to `.env.tpl` if that is not done already (I did not see it in this PR but it might have gone in already?)
2. Remove the bit that prevents the email from sending within localstack (unless there's a problem with that in practice that you've already seen!)

The rest of my comments are just some thoughts about it, but as a proof of concept this seems to work and can be iterated on further!

services/app-api/libs/localstack.ts
rocio-desantiago: ooope nope. I forgot I removed that

deployment/local/prerequisites.ts
rocio-desantiago: Thank youuuu

deployment/stacks/api.ts
rocio-desantiago: Applied ✅

services/app-api/libs/launchdarkly-lib.ts
rocio-desantiago: Applied ✅ Still works 🙌

services/app-api/libs/tests/launchdarkly-lib.test.ts
rocio-desantiago: expect.anything() works for me 😄

services/app-api/utils/notifications/email.ts
rocio-desantiago: updated and correct since `VerifyEmailIdentityCommand` was added to local setup and we are verified ✅

stacy-coforma (approved):
looks like it's working well! we can automate the localstack validation of the sender email for better local setup/testing, but that'll have to go into `core` I think, which may be helpful for RHTP as well (and any others we add email to in the future) 🙂

PR #668, opened 2026-05-04
Add Skip to main link

sopspops:
### Description

* There's no link to skip to main content, so add skip to main link

### Related ticket(s)

CMDCT-5107

---
### How to test

 https://d1n5i8aysc156h.cloudfront.net/
* Navigate through the app
* Check the first thing being focused is Skip to main link

### Notes

---
### Pre-review checklist

---
#### Review

#### Security


cy-by (dismissed):
## ✅ Skip nav

Works as I expected.

## Dev login nit pick

This isn't blocked the PR from merging.

On the [deployed env](https://d1n5i8aysc156h.cloudfront.net/), I saw that there's more space between the header and the form. I think this is coming from `flex` being applied to the `<main>`.

Here's a video showing the different on the dev login (other pages look ok):

https://github.com/user-attachments/assets/aba82a8e-c8a7-4794-a00e-809c18063432

Here's the `<main>` CSS in dev:

Image

Here's the `<main>` CSS in the deployed env:

Image

services/ui-src/src/components/skipNav/SkipNav.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+describe("<SkipNav />", () => {
+  test("renders skip link with accessible text and becomes focusable", async () => {
+    render(skipNavComponent);
+
+    const skipNavLink = screen.getByRole("link", {
+      name: "Skip to main content",
+    });
+
+    skipNavLink.focus();
```
stacy-coforma: Bit of a testing nit - setting focus here and then checking focus in a couple of lines will not simulate a proper `tab` from the user to test whether the focus is properly assigned with the keyboard action.  We could try something like this instead:

```suggestion
    await userEvent.tab();
```

which would require importing the `@testing-library/user-event` functionality, but I think would be more reflective of user behavior and a more valuable test 👍 

(Note: I am not sure if that needs to be wrapped in `act` or not for the assertion to pick it up 😅)

stacy-coforma (commented):
Looks good and works well!  I left one annoying comment in the unit test 😅

cy-by (approved):
My comment has been addressed

stacy-coforma (approved):

PR #672, opened 2026-05-07
Package Upgrades, various

benmartin-coforma:
### Description
I believe this will resolve all of the `Security and quality` warnings currently outstanding on the repo.

When possible, I resolved warnings by upgrading our direct dependencies. In other cases, I found that overriding versions for our transient dependencies was more practical. In one case (`uuid`), I just dismissed the warning as a non-issue.

Almost all changes are minor/patch version bumps, with no code changes required in our repo. The exceptions are:
- Upgrading `vite` 5 -> 8, which also meant removing `vite-tsconfig-paths` (it's built in now) and upgrading `@vitejs/plugin-react` 4 -> 6.
- `http-proxy-agent` 5 -> 6. This is a transient dependency: via jsdom, via jest. The warning it resolves is actually on `@tootallnate/once`; the old version of `http-proxy-agent` depended on it and the new one does not.
  - I would have preferred to upgrade jest, the direct dependency in this case. And that would have resolved the warning. But that would have meant moving to jsdom 26, which would have meant breaking all of our unit tests that mock window.location. See https://github.com/jsdom/jsdom, issue 3492.
  - We _should_ upgrade jest soon, and deal with the window.location fallout. It affects a handful of tests, but only a handful.
  - Will it cause problems in the meantime, for us to give jsdom the "wrong" version of http-proxy-agent? No. It's possible that there are breaking changes, and jsdom would misbehave... _if_ we did anything at all with http in our unit tests. We don't, so I am confident any potentially-fraught code paths will be completely avoided.
- `@types/dompurify` 2 -> 3, even though `dompurify` was already at 3. This didn't resolve a warning, I just don't like having the types package out of sync.

### Related ticket(s)
CMDCT-5997

---
### How to test

If the tests pass and the site is up, I think we're good.

### Notes
Here are the notes I took while working on this. They're probably not valuable at this point, but here they are anyway.
- Handlebars.js JS injection via pre-parsed AST
  - non-issue; we only run handlebars to parse our own test code
  - resolved by upgrading ts-jest 29.2.5 -> 29.4.9
  - Why do we have ts-jest? Apparently it's only in the API; the UI doesn't have it and its tests run just fine.
    - Apparently it's either ts-jest, or @babel/preset-typescript. Let's keep it as-is for now.
    - Someday, vitest.
- tar Race Condition with Unicode Ligatures
  - We have tar@7.5.2 due to node-gyp@12.1.0
  - Addressed in tar@7.5.4, which was picked up by node-gyp@12.2.0
  - We have node-gyp@12.1.0 due to fsevents, @parcel/watcher, and node-addon-api
    - We have @parcel/watcher and node-addon-api due to sass
      - @parcel/watcher no longer depends on node-gyp, and neither does node-addon-api
      - We are already using @parcel/watcher@2.5.6, which already dropped that dep
        - but we _ask_ for @parcel/watcher@^2.4.1, which presumably still has it
      - According to NPM, sass 1.99.0 does not depend on @parcel-watcher. According to yarn just now, it does.
    - We have fsevents (various patches) due to various tools: jest, tsx, playwright, and vite
  - This all makes it look like updating our actual dependencies away from the issue is a no-go.
  - Resolved by resolution.
- minimatch reDoS
  - We have minimatch due to aws-cdk-lib, test-exclude (due to jest), and glob (for all kinds of dev reasons)
  - We've seen this in QMR and SEDS. The answer is going to be resolutions.
  - Going to latest patch: 10.2.5, 3.1.5, 6.2.3
    - Except aws-cdk-lib asks for 10.2.3 already, which afaik is safe. Leaving that as-is.
- brace-expansion DoS
  - Only required via minimatch
  - Already resolved by minimatch overrides.
- tar arbitrary file creation
  - We already know we've got to override this, because we did so yesterday.
  - Resolved by bumping that override a little further.
- Handlebars JS injection via unescaped names
  - Already resolved by prior override.
- tar symlink path traversal
  - lmao, bump it again
- tar hardlink path traversal
  - already resolved lmao
- tar arbitrary file creation again
  - already resolved
- lodash code injection
  - We have lodash due to CMSDS, which doesn't seem to have a patch available
  - Override it is, then.
- There are SIX MORE HANDLEBARS vulns.
  - All of them patched by 4.7.9, which we already have thanks to the ts-jest upgrade.
- semver reDoS
  - Fixed in 7.5.2, 6.3.1, 5.7.3
  - We have so many copies, for so many reasons
  - Ignoring the safe versions:
    - 6.3.0 from istanbul-lib-instrument
    - 7.3.5 from (ultimately) node-gyp 12.1.0
    - 7.3.5 from pa11y
  - And even the latest node-gyp still pulls in 7.3.5, so upgrades can't help us.
  - Overridden.
- path-to-regexp reDoS
  - Fixed in 8.4.0
  - included, indirectly, by aws-sdk-client-mock
    - It's not too bad to homebrew our own aws-sdk-client-mocks with vitest
    - Later let's uninstall this package
  - For now, overridden.
- picomatch reDoS (x2)
  - Fixed in 2.3.2, 3.0.2, 4.0.4
  - We have so many.
  - Overridden.
- Vite file traversal (x2)
  - Upgraded to v8, why not
- dompurify XSS holes (x3)
  - Upgraded to 3.4
- lodash prototype pollution
  - Already resolved
- brace-expansion DoS (x2)
  - Fixed in 1.1.13, 5.0.5
  - We have 5.0.2 via minimatch 10.2.4, but can bump to 10.2.5 to get 5.0.5
  - But we have 1.1.7 via minimatch 3.1.5, which is already latest v3
  - Overrode minimatch 10 & b-e 1
- UUID buffer shenanigans
  - Dismissed alert. We should be able to remove the LD SDK tbh.
    - Removed LD SDK, committed to a branch, will deal with that later.
- PostCSS XSS
  - Already fixed by Vite upgrade
- fast-xml-parser CDATA injection
  - Fixed in 5.7.0
  - We have 5.3.6 via amplify and 5.5.6 via aws-sdk
  - Bumping those packages gets us the fix!
- fast-xml-parser Entity Expansion Limits
  - Fixed by prior bumps
- ajv reDoS
  - Fixed in 8.18.0
  - We have 8.0.1, via table@6.9.0, via aws-cdk-lib@2.243.0
    - table@6.9.0 is latest, so nothing we can do there
  - Overridden.
- esbuild dev server exposure
  - Already resolved by vite upgrade
- yaml stackoverflow
  - Fixed in 1.10.3
  - We have 1.10.2 via aws-cdk-lib, which we can bump away from
  - We have 1.10.0 via cosmiconfig@7.0.0,
    via babel-plugin-macros@3.1.0,
    via @emotion/babel-plugin@11.11.4,
    via @emotion/react@11.11.4 and @emotion/styled@11.13.5
    - And getting latest on these doesn't touch babel-plugin-macros,
      so we're cooked.
  - Upgraded aws-cdk-lib, overrode the other.
- @tootallnate/once hanging promise
  - We have once 2 via http-proxy-agent@5.0.0,
    via jsdom@20.0.0,
    via jest-environment-jsdom@29.7.0
  - Fixed in once 3.0.1
  - http-proxy-agent
    v9 does not have the once dep
    v8 also does not
    v7 does not
    v6 does not
    v5 does not??? What is going on
  - jsdom is 9 major versions behind.
    v29 does not depend on http-proxy-agent
    v28 does not
    v27 depends on http-proxy-agent@7.0.2
    v24 depends on http-proxy-agent@7.0.0
    Gotta go back to 22.1.1 to get http-proxy-agent@5.0.0
  - jest-environment-jsdom is only 1 major version behind
    30.3.0 -> jsdom 26.1.0, which is plenty far enough
    30.0.0-beta -> 26.0.0, also plenty
    30.0.0-alpha -> 22.0.0, not enough.
  - So we can upgrade away from this by bumping jest to 30.
  - Might as well go latest, 30.3.0
  - And take all its friends along
    - jest, @types/jest
    - jsdom, @types/jsdom
    - jest-axe
    - jest-environment-jsdom
  - Unfortunately, moving jsdom to 26 breaks all of our window.location tests
  - So never mind that.
  - Overriding http-proxy-agent from 5 to 6.
  - Someday we will have to bite the jsdom@26 bullet, but not today.

---
### Pre-review checklist

---
#### Review
- ~[ ] Design: This work has been reviewed and approved by design, if necessary~
- ~[ ] Product: This work has been reviewed and approved by product owner, if necessary~

#### Security

- ~[ ] These changes are significant enough to require an update to the SIA.~
- ~[ ] These changes are significant enough to require a penetration test.~


stacy-coforma (approved):
thank you for working out the local issues with me as well 🙏

sopspops (approved):
weee thank you ben!

PR #673, opened 2026-05-07
Dashboard table has extra empty column header

sopspops:
### Description

* When expecting headers on dashboard (for example, Submission name), there's an extra empty <th> at the end.
* Removed Actions from the header and push it at the end, so it doesn't push an empty <th>

### Related ticket(s)

CMDCT-5984

---
### How to test

https://d3gl5jaimgoyuj.cloudfront.net/
* Navigate to any report
* Expect dashboard header and see there's no <th> with empty string

### Notes

---
### Pre-review checklist

---
#### Review

#### Security

### DIFF
--- a/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
@@ -267,16 +267,17 @@ export const DashboardTable = ({
   const showAdminControlsColumn = userIsAdmin;
 
   // Build header columns based on defined behaviors per role
-  const headers = [
+  const baseHeaders = [
     "Submission name",
     "Reporting year",
     "Last edited",
     "Edited by",
     "Status",
-    "Actions",
   ];
+  const headers = [...baseHeaders];
   if (showReportSubmissionsColumn) headers.push("#");
-  headers.push("");
+  headers.push("Actions");
+  if (showAdminControlsColumn) headers.push("", "");
   const { reportType } = useParams();
   const tableContent = {
     caption: `${getReportName(reportType)}s`,
--- a/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
+++ b/services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
@@ -267,14 +267,14 @@ export const DashboardTable = ({
   const showAdminControlsColumn = userIsAdmin;
 
   // Build header columns based on defined behaviors per role
-  const baseHeaders = [
+  const headers = [
     "Submission name",
     "Reporting year",
     "Last edited",
     "Edited by",
     "Status",
   ];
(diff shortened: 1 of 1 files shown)


services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -267,16 +267,17 @@ export const DashboardTable = ({
   const showAdminControlsColumn = userIsAdmin;
 
   // Build header columns based on defined behaviors per role
-  const headers = [
+  const baseHeaders = [
```
stacy-coforma: I don't think we need this variable; we reference it here and then it doesn't look like it's used again, so I think it would be fine to just define `headers` here.

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -267,16 +267,17 @@ export const DashboardTable = ({
...
     "Last edited",
     "Edited by",
     "Status",
-    "Actions",
   ];
+  const headers = [...baseHeaders];
   if (showReportSubmissionsColumn) headers.push("#");
-  headers.push("");
+  headers.push("Actions");
+  if (showAdminControlsColumn) headers.push("", "");
```
stacy-coforma, line 280: I did not realize that an admin user would need two extra columns, which are both currently without headers 😕 we should circle up with Cypress and somebody in design at some point to see how we need to address this, but it's a separate problem from the one you are solving in this ticket.

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
```diff
@@ -267,16 +267,17 @@ export const DashboardTable = ({
...
     "Reporting year",
     "Last edited",
     "Edited by",
     "Status",
-    "Actions",
   ];
+  const headers = [...baseHeaders];
   if (showReportSubmissionsColumn) headers.push("#");
-  headers.push("");
+  headers.push("Actions");
```
stacy-coforma, line 279: good catch on the ordering

stacy-coforma (changes requested):
I left so many comments for like a 5 line change 😅 but just the one is actionable right now!  Let's talk to Cypress and Amanda about those empty column headings for admins after CfA 👀 just recording that as an observation for followup

services/ui-src/src/components/pages/Dashboard/DashboardTable.tsx
sopspops: ooooh yup yup!! thank you Stacy!!

stacy-coforma (approved):
tested with a few different user types and the table does appear correctly for the state user; already noted that admin/business users with the extra columns will need to have those empty headers addressed, but we can do that separately once everyone has time to review post-CfA.

PR #678, opened 2026-05-20
Update NDR rename script

benmartin-coforma:
### Description
Wait until after everything is defined to invoke `main()`


ntsummers1 (approved):

stacy-coforma (approved):
confirmed, I did have to do this when I ran it locally, good edit

